PR #2, opened 2020-11-09
Install deps

cSparkle:
This adds in all the deps and devDeps we're likely to need going forward, based on the Package.json from the React Native SDK.


joshdholtz (approved):
🔥

PR #3, opened 2020-11-09
Setup client structure

cSparkle:
Setup the general boilerplate structure for the client, backend, error handlers, and interfaces needed to begin writing CAP plugin methods. Much of this was abstracted from the [React Native](https://github.com/getsentry/sentry-react-native) and [Cordova projects](https://github.com/getsentry/sentry-cordova)


src/backend.ts
```diff
@@ -0,0 +1,101 @@
...
+
+    const scope = getCurrentHub().getScope();
+    if (scope) {
+      scope.addScopeListener(internalScope => {
+        this._nativeCall(
+          'setExtraContext',
+          (internalScope as any)._extra,
+        ).catch(() => {
+          // We do nothing since scope is handled and attached to the event.
+          // This only applies to android.
```
joshdholtz, line 67: I think we can delete this line since this should apply to both iOS and Android (since iOS will have envelopes soon)?

src/backend.ts
```diff
@@ -0,0 +1,101 @@
...
+        });
+        this._nativeCall('setTagsContext', (internalScope as any)._tags).catch(
+          () => {
+            // We do nothing since scope is handled and attached to the event.
+            // This only applies to android.
```
joshdholtz, line 72: I think we can delete this line since this should apply to both iOS and Android (since iOS will have envelopes soon)?

src/backend.ts
```diff
@@ -0,0 +1,101 @@
...
+          },
+        );
+        this._nativeCall('setUserContext', (internalScope as any)._user).catch(
+          () => {
+            // We do nothing since scope is handled and attached to the event.
+            // This only applies to android.
```
joshdholtz, line 78: I think we can delete this line since this should apply to both iOS and Android (since iOS will have envelopes soon)?

src/backend.ts
```diff
@@ -0,0 +1,101 @@
...
+          },
+        );
+        this._nativeCall(
+          'addBreadcrumb',
+          (internalScope as any)._breadcrumbs.pop(),
+        ).catch(() => {
+          // We do nothing since scope is handled and attached to the event.
+          // This only applies to android.
```
joshdholtz, line 86: I think we can delete this line since this should apply to both iOS and Android (since iOS will have envelopes soon)?

src/wrapper.ts
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    if (this.platform !== 'ios') {
+      // Only ios uses deviceContexts, return an empty object.
+      return {};
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+    return SentryCapacitor.deviceContexts();
+  },
```
joshdholtz, line 67: @bruno-garcia This isn’t needed once we transition the iOS SDK to using envelopes, right? 🤔

joshdholtz (commented):
A few thoughts/comments/questions otherwise looks like a good start to me 😊

cSparkle:
@joshdholtz I updated things based on your comments!

joshdholtz (approved):
This looks good to me!

PR #4, opened 2020-11-12
Android client setup

cSparkle:
Lots more setup here of boilerplate, middleware JS functions, and some updates to the CapacitorOptions interface. 

The "Init" function is working, but needs tests which I'll work on in the next PR.

### DIFF
--- a/src/backend.ts
+++ b/src/backend.ts
@@ -3,12 +3,8 @@ import { BrowserBackend } from '@sentry/browser/dist/backend';
 import { BaseBackend } from '@sentry/core';
 import { getGlobalObject } from '@sentry/utils';
 
-import { Capacitor, Plugins } from '@capacitor/core';
-
 import { CapacitorOptions } from './options';
 import { NATIVE } from './wrapper';
-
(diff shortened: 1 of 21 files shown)


src/backend.ts
```diff
@@ -25,18 +22,62 @@ export class CapacitorBackend extends BaseBackend<BrowserOptions> {
...
   /**
    * Has Capacitor on window?
    */
-  private _isCapacitor(): boolean {
+  // private _isCapacitor(): boolean {
+  //   return (
+  //     getGlobalObject<any>().capacitor !== undefined ||
+  //     getGlobalObject<any>().Capacitor !== undefined
+  //   );
+  // }
```
joshdholtz, line 71: Do we not need this at all? Or are we saving this for later?

joshdholtz (approved):
One question otherwise 👍

src/backend.ts
cSparkle: @joshdholtz I'm potentially saving this for later. I'm uncertain right now whether I'll need it yet.

PR #6, opened 2020-11-13
chore: add in correct eslint config files and lint everything

cSparkle:
Realized I wasn't using the proper Sentry eslint config, so add these in and linted everything.


joshdholtz (approved):
👍

PR #7, opened 2020-11-16
Fix error handling

cSparkle:
Experienced errors in build due to the unknown ErrorUtils imports in the CapacitorErrorHandlers.js file. This file was copied from the React-Native repo, and it doesn't appear Capacitor has any sort of comparable ErrorUtils interface. 

Considering that Cordova doesn't utilized this custom error handlers integration, I'm removing for now as it doesn't appear to be necessary.


joshdholtz (approved):
👍

PR #8, opened 2020-11-16
Capture envelope

cSparkle:
Adds captureEnvelope functionality in the main .java plugin file.

### DIFF
--- a/android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
+++ b/android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
@@ -142,12 +142,15 @@ public class SentryCapacitor extends Plugin {
             File installation =  new File(sentryOptions.getOutboxPath(), UUID.randomUUID().toString());
             try (FileOutputStream out = new FileOutputStream(installation)) {
                 out.write(envelope.getBytes(Charset.forName("UTF-8")));
+            } catch (Exception e) {
+                logger.info("Error writing envelope: ", e);
             }
         }
         catch (Exception e) {
-            logger.info("Error reading envelope.");
+            logger.info("Error reading envelope: ", e);
         }
 
+        logger.info("Successfully captured envelope")
         call.resolve();
     }
 }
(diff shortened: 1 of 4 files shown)


android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
```diff
@@ -129,4 +134,20 @@ public void startWithOptions(final PluginCall capOptions) {
...
     }
+
+    @PluginMethod
+    public void captureEnvelope(PluginCall call) {
+        String envelope = call.getString("envelope");
+        try {
+            File installation =  new File(sentryOptions.getOutboxPath(), UUID.randomUUID().toString());
+            try (FileOutputStream out = new FileOutputStream(installation)) {
+                out.write(envelope.getBytes(Charset.forName("UTF-8")));
+            }
```
joshdholtz, line 145: 1) Do we want to catch and log an error here if writing fails?
2) Is there any debug logger that we want here for when its successfully written?

joshdholtz (commented):
Just a two small questions!

android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
cSparkle: @joshdholtz  Updated!

android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
```diff
@@ -129,4 +134,23 @@ public void startWithOptions(final PluginCall capOptions) {
...
+                out.write(envelope.getBytes(Charset.forName("UTF-8")));
+            } catch (Exception e) {
+                logger.info("Error writing envelope: ", e);
+            }
+        }
+        catch (Exception e) {
+            logger.info("Error reading envelope: ", e);
+        }
+
+        logger.info("Successfully captured envelope")
```
joshdholtz, line 153: This will get called when the exceptions are caught 😊  You might want to move this up to right after the `out.write()`

joshdholtz (changes requested):
One more small change!

android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
cSparkle: oh, good point. On it!

cSparkle: @joshdholtz done!

joshdholtz (approved):
🔥 LGTM

PR #9, opened 2020-11-17
fix: resolve test issues from changing how native availability is checked

cSparkle:
Tests broke when I changed how I checked whether the plugin was available, so this fixes what was broken.

Screen Shot 2020-11-17 at 11 23 48 AM


joshdholtz (approved):
🔥 LGTM

PR #10, opened 2020-11-18
fix: updated type of data received to fix captureenvelope and getstringbyteslength android methods

cSparkle:
This implements the captureMessage and captureException functionality in android and in the JS middleware. There are also fixes to the tests and updates to the type of data being sent to captureEnvelope and getStringBytesLength (the wrong data was previously being sent).

Testing of both the captureException and captureMessage methods was used in a project I created on the Sentry website, and both events were successfully sent: 
Screen Shot 2020-11-18 at 1 22 44 PM

### DIFF
--- a/android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
+++ b/android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
@@ -46,11 +46,12 @@ public class SentryCapacitor extends Plugin {
 
         if (this.context == null) {
             this.context = this.bridge.getContext();
-            try {
-                this.packageInfo = this.context.getPackageManager().getPackageInfo(this.getContext().getPackageName(), 0);
-            } catch (Exception e) {
-                logger.info("Error getting package info.");
-            }
+        }
+
+        try {
+            this.packageInfo = this.context.getPackageManager().getPackageInfo(this.getContext().getPackageName(), 0);
+        } catch (Exception e) {
+            logger.info("Error getting package info.");
         }
     }
 
@@ -95,19 +96,6 @@ public class SentryCapacitor extends Plugin {
 
                 options.setBeforeSend(
                     (event, hint) -> {
-                        // TODO Need to look into whether this is needed for Capacitor
-                        // React native internally throws a JavascriptException
-                        // Since we catch it before that, we don't want to send this one
-                        // because we would send it twice
-                        // try {
-                        //     SentryException ex = event.getExceptions().get(0);
-                        //     if (null != ex && ex.getType().contains("JavascriptException")) {
-                        //         return null;
-                        //     }
-                        // } catch (Exception e) {
-                        //     // We do nothing
-                        // }
-
                         // Add on the correct event.origin tag.
                         // it needs to be here so we can determine where it originated from.
                         SdkVersion sdkVersion = event.getSdk();
@@ -199,11 +187,13 @@ public class SentryCapacitor extends Plugin {
             } catch (Exception e) {
                 logger.info("Error writing envelope.");
                 call.reject(String.valueOf(e));
+                return;
             }
         }
         catch (Exception e) {
             logger.info("Error reading envelope.");
             call.reject(String.valueOf(e));
+            return;
         }
     }
     
@@ -218,6 +208,8 @@ public class SentryCapacitor extends Plugin {
             } catch (UnsupportedEncodingException e) {
                 call.reject(String.valueOf(e));
             }
+        } else {
+            call.reject("Could not calculate string length.");
         }
     }
 
@@ -289,6 +281,7 @@ public class SentryCapacitor extends Plugin {
         Sentry.configureScope(scope -> {
             scope.setExtra(key, extra);
         });
+        call.resolve();
     }
 
     @PluginMethod
@@ -298,5 +291,6 @@ public class SentryCapacitor extends Plugin {
         Sentry.configureScope(scope -> {
             scope.setTag(key, value);
         });
+        call.resolve();
     }
 }
--- a/android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
+++ b/android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
@@ -1,40 +1,58 @@
 package io.sentry.capacitor;
 
+import com.getcapacitor.JSArray;
+import com.getcapacitor.JSObject;
 import com.getcapacitor.NativePlugin;
 import com.getcapacitor.Plugin;
 import com.getcapacitor.PluginCall;
 import com.getcapacitor.PluginMethod;
 
+import io.sentry.Breadcrumb;
 import io.sentry.Integration;
+import io.sentry.Sentry;
+import io.sentry.SentryLevel;
 import io.sentry.SentryOptions;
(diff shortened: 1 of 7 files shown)


android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
```diff
@@ -81,14 +99,14 @@ public void startWithOptions(final PluginCall capOptions) {
...
-                            // We do nothing
-                        }
+                        // try {
+                        //     SentryException ex = event.getExceptions().get(0);
+                        //     if (null != ex && ex.getType().contains("JavascriptException")) {
+                        //         return null;
+                        //     }
+                        // } catch (Exception e) {
+                        //     // We do nothing
+                        // }
```
joshdholtz, line 109: Will this get uncommented in a later version? Or can we delete this?

android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
```diff
@@ -1,40 +1,58 @@
...
+        super.load();
 
-    //     JSObject ret = new JSObject();
-    //     ret.put("value", value);
-    //     call.success(ret);
-    // }
+        if (this.context == null) {
+            this.context = this.bridge.getContext();
+            try {
+                this.packageInfo = this.context.getPackageManager().getPackageInfo(this.getContext().getPackageName(), 0);
```
joshdholtz, line 50: Does this `packInfo` not need to get set if `context` already exists?

android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
```diff
@@ -132,7 +150,37 @@ public void startWithOptions(final PluginCall capOptions) {
             }
         );
 
-        capOptions.resolve();
+        JSObject resp = new JSObject();
+        resp.put("value", true);
+        capOptions.resolve(resp);
```
joshdholtz, line 155: What is this `"value"` here for? 🤔

android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
```diff
@@ -142,15 +190,113 @@ public void captureEnvelope(PluginCall call) {
...
+                logger.info("Successfully captured envelope.");
+
+                JSObject resp = new JSObject();
+                resp.put("value", envelope);
+                call.resolve(resp);
+
             } catch (Exception e) {
-                logger.info("Error writing envelope: ", e);
+                logger.info("Error writing envelope.");
+                call.reject(String.valueOf(e));
```
joshdholtz, line 201: Might want to do a `return` after this just so its clear that nothing should happen outside of this try

android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
```diff
@@ -142,15 +190,113 @@ public void captureEnvelope(PluginCall call) {
...
             }
         }
         catch (Exception e) {
-            logger.info("Error reading envelope: ", e);
+            logger.info("Error reading envelope.");
+            call.reject(String.valueOf(e));
```
joshdholtz, line 206: Might want to do a `return` after this just so its clear that nothing should happen outside of this try

android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
```diff
@@ -142,15 +190,113 @@ public void captureEnvelope(PluginCall call) {
...
+        if (call.getData().has("string")) {
+            String payload = call.getString("string");
+            try {
+                JSObject resp = new JSObject();
+                resp.put("value", payload.getBytes("UTF-8").length);
+                call.resolve(resp);
+            } catch (UnsupportedEncodingException e) {
+                call.reject(String.valueOf(e));
+            }
+        }
```
joshdholtz, line 221: Does `call` need to get resolved in the `else` case here?

android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
```diff
@@ -142,15 +190,113 @@ public void captureEnvelope(PluginCall call) {
...
+        });
+    }
+
+    @PluginMethod
+    public void setExtra(PluginCall call) {
+        String key = call.getString("key");
+        String extra = call.getString("extra");
+        Sentry.configureScope(scope -> {
+            scope.setExtra(key, extra);
+        });
```
joshdholtz, line 291: Does `call` need to get resolved?

android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
```diff
@@ -142,15 +190,113 @@ public void captureEnvelope(PluginCall call) {
...
+    }
+
+    @PluginMethod
+    public void setTag(PluginCall call) {
+        String key = call.getString("key");
+        String value = call.getString("value");
+        Sentry.configureScope(scope -> {
+            scope.setTag(key, value);
```
joshdholtz, line 299: Does `call` need to get resolved?

src/web.ts
```diff
@@ -27,8 +27,10 @@ export class SentryCapacitorWeb
   /**
    *
    */
-  captureEnvelope(): Promise<Response> {
+  captureEnvelope(payload: { envelope: string }): Promise<Response> {
     // TODO integrate web
+    /* eslint-disable-next-line no-console */
+    console.log(payload.envelope);
```
joshdholtz, line 33: 💣  this `console.log`?

src/web.ts
```diff
@@ -67,11 +69,13 @@ export class SentryCapacitorWeb
...
    */
-  getStringBytesLength(payloadString: 'somepayload'): number {
+  getStringBytesLength(payload: {
+    string: string;
+  }): Promise<{ value: number }> {
     // TODO integrate web
     /* eslint-disable-next-line no-console */
-    console.log(payloadString);
-    return 12;
+    console.log(payload.string);
```
joshdholtz, line 77: Same here. 💣 this?

joshdholtz (changes requested):
Left some questions/changes and stuff. But otherwise this is 🔥

android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
cSparkle: Capacitor only allows promises to return their own JSObject type instead of a string, so I unfortunately can't resolve with just "true".

src/web.ts
cSparkle: @josh I was just using these temporarily to please the linter, but I can add an eslint exception instead if you'd prefer!

android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
cSparkle: good catch!

android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
cSparkle: ahh, good catch. I'll move it outside that if statement so packageInfo is set at all times.

joshdholtz (approved):
🔥

bruno-garcia (approved):
👍

PR #11, opened 2020-11-18
feat: finish up last bits of native crash functionality

cSparkle:
Turns out native crash functionality was mostly complete, so I just finished up that last bits like adding in a test and ensuring the native crash method communicates with the SDK


joshdholtz (approved):
Ha! So easy 😊 Love it!

PR #12, opened 2020-11-19
feat: add setUser functionality

cSparkle:
Adapts the setUser functionality from the React-Native project.

### DIFF
--- a/android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
+++ b/android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
@@ -17,6 +17,7 @@ import io.sentry.android.core.AnrIntegration;
 import io.sentry.android.core.NdkIntegration;
 import io.sentry.android.core.SentryAndroid;
 import io.sentry.protocol.SdkVersion;
+import io.sentry.protocol.User;
 
 import java.io.File;
 import java.io.FileOutputStream;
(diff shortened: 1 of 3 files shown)


android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
```diff
@@ -162,6 +163,50 @@ private Level logLevel(int level) {
...
+                    }
+
+                    if (user.has("username")) {
+                        userInstance.setUsername(user.getString("username"));
+                    }
+
+                    if (user.has("ip_address")) {
+                        userInstance.setIpAddress(user.getString("ip_address"));
+                    }
+                }
```
joshdholtz, line 191: Do we need an `else` here to clear out the user if nothing?

android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
```diff
@@ -162,6 +163,50 @@ private Level logLevel(int level) {
...
+                    Map<String, String> otherUserKeysMap = (Map<String, String>) call.getObject("otherUserKeys");
+
+                    for (Map.Entry<String, String> entry: otherUserKeysMap.entrySet()) {
+                        String key = entry.getKey();
+                        String value = entry.getValue();
+                        otherUserKeysMap.put(key, value);
+                    }
+
+                    userInstance.setOthers(otherUserKeysMap);
+                }
```
joshdholtz, line 203: Same here. Do we need to clear this if not set?

joshdholtz (changes requested):
Two questions!

android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
cSparkle: good catch!

cSparkle: @joshdholtz, actually I think the enclosing if statement should take care of this? It checks whether both the "user" and "otherUserKeys" is available, and if not, sets the scope to null.

joshdholtz: The enclosing if will clear if both are not set. What if only one of them is available? Is there no way to clear just one? I think that is my concern. I don't know how the other SDKs handle that 🤷‍♂️

cSparkle: @joshdholtz this is how the React Native SDK was setup. I think the idea is that as long as one of the arguments ("user" or "otherUserKeys") is there, it can set that as the userInstance. Mostly likely it seems "user" will be the argument being sent and "otherUserKeys" is optional.

@bruno-garcia perhaps you can provide some insight? I'm less familiar with the "user" and "otherUserKeys" data and how they are eventually used by the Sentry API.

PR #15, opened 2020-11-30
Sample project

cSparkle:
This update includes an Ionic/Angular example project that mimics the exception and message call found in the Sentry Dart and Browser examples. I created an example folder in case it made sense to include examples using other frameworks. 

**Important** I uncovered a bug related to incorrectly casting types in the Android implementation while building the sample app that I'm planning to resolve in the next PR. The sample app works and logs exceptions/messages correctly, but does not properly add breadcrumbs or custom tags atm.


example/ionic-angular/README.md
```diff
@@ -0,0 +1,69 @@
...
+## Add Capacitor to an existing Ionic project
+Full details can be found at capacitorjs.com/docs/getting-started/with-ionic
+1. `ionic integrations enable capacitor`
+2. `npx cap init [appName] [appId]`
+3. `ionic build`
+4. `npx cap add ios`
+5. `npx cap add android`
+6. `npx cap sync`
+
+## Add Sentry/Capacitor to your project
```
bruno-garcia, line 27: Is this relevant to the sample? Or should this just go under the docs for general instructions on how to add it to an app?

example/ionic-angular/android/app/src/androidTest/java/com/getcapacitor/myapp/ExampleInstrumentedTest.java
```diff
@@ -0,0 +1,27 @@
+package com.getcapacitor.myapp;
```
bruno-garcia, line 1: If you don't plan to have these instrumented tests on the example app, maybe we can remove them to have less files around

example/ionic-angular/e2e/protractor.conf.js
```diff
@@ -0,0 +1,36 @@
+// @ts-check
+// Protractor configuration file, see link for more information
+// https://github.com/angular/protractor/blob/master/lib/config.ts
+
+const { SpecReporter, StacktraceOption } = require('jasmine-spec-reporter');
```
bruno-garcia, line 5: Also: Unless we're doing some e2e tests (react native project does that), we don't need this file?

Happy to keep it if we're using though.

example/ionic-angular/src/app/tab1/tab1.page.spec.ts
```diff
@@ -0,0 +1,25 @@
+import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+import { IonicModule } from '@ionic/angular';
+import { ExploreContainerComponentModule } from '../explore-container/explore-container.module';
+
+import { Tab1Page } from './tab1.page';
+
+describe('Tab1Page', () => {
```
bruno-garcia, line 7: Maybe we get rid of these, unless they actually test something and we're running it in CI

bruno-garcia (approved):
LGTM! Left some notes for improving though

PR #17, opened 2020-12-02
add try catch for logger, small updates to example app

cSparkle:
I was originally pulling the wrong type from the PluginCall for the setTag and breadcrumbs methods. I updated the typecasting to the appropriate types and made some small updates to the verbiage in the example app to make the messages clearer.

### DIFF
--- a/android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
+++ b/android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
@@ -1,11 +1,11 @@
 package io.sentry.capacitor;
 
-import com.getcapacitor.JSArray;
 import com.getcapacitor.JSObject;
 import com.getcapacitor.NativePlugin;
 import com.getcapacitor.Plugin;
 import com.getcapacitor.PluginCall;
 import com.getcapacitor.PluginMethod;
+import com.google.gson.Gson;
 
 import io.sentry.Breadcrumb;
 import io.sentry.Integration;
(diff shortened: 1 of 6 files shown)


android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
```diff
@@ -326,21 +333,27 @@ public void clearBreadcrumbs() {
...
-        call.resolve();
+        if (call.getData().has("key") && call.getData().has("extra")) {
+            Sentry.configureScope(scope -> {
+                String key = call.getString("key");
+                String extra = call.getString("extra");
+                scope.setExtra(key, extra);
+            });
+
+            call.resolve();
+        }
```
joshdholtz, line 344: Does `call.resolve()` need to be executed outside of the if statement?

android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
```diff
@@ -326,21 +333,27 @@ public void clearBreadcrumbs() {
...
-        call.resolve();
+        if (call.getData().has("key") && call.getData().has("value")) {
+            Sentry.configureScope(scope -> {
+                String key = call.getString("key");
+                String value = call.getString("value");
+                scope.setTag(key, value);
+            });
+
+            call.resolve();
+        }
```
joshdholtz, line 357: Same here! Does this `call.resolve()` need to be executed outside of the if statement?

joshdholtz (changes requested):
Two small things! Otherwise 👍

android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
cSparkle: good catch!

PR #19, opened 2020-12-23
finish up startwithoptions

cSparkle:
Init function setup for ios. No testing done yet as it will be more feasible once the entire iOS swift controller is setup.


ios/Plugin/Plugin.swift
```diff
@@ -1,17 +1,46 @@
...
-        call.success([
-            "value": value
-        ])
+    
+    @objc public func startWithOptions(capOptions: CAPPluginCall) {
+        SentrySDK.start { options in
+            options.dsn = capOptions.getString("dsn") ?? ""
+            options.debug = capOptions.getBool("debug") ?? false
+            options.environment = capOptions.getString("environment") ?? "production"
+            options.enableAutoSessionTracking = capOptions.getBool("enableAutoSessionTracking") ?? false
```
bruno-garcia, line 13: The default value for this field is `true` but to avoid duplicating it here I suggest checking whether a value existed in the map and only then mutating the `options` object. This way we're free to change the default values in the iOS SDK without having to remember/change here too.

bruno-garcia:
I just realized something. Ideally here we can use the overload of the cocoa SDK that takes a Dictionary, so it can deal with looking up the keys and apply, etc. `startWithOptionsObject`.

https://github.com/getsentry/sentry-react-native/blob/master/ios/RNSentry.m#L67

ios/Plugin/Plugin.swift
```diff
@@ -1,17 +1,51 @@
...
 @objc(SentryCapacitor)
 public class SentryCapacitor: CAPPlugin {
-
-    @objc func echo(_ call: CAPPluginCall) {
-        let value = call.getString("value") ?? ""
-        call.success([
-            "value": value
-        ])
+    
+    @objc public func startWithOptions(capOptions: CAPPluginCall) {
```
cSparkle, line 8: @bruno-garcia is this what you had in mind?

bruno-garcia: yeah! awesome

ios/Plugin/Plugin.swift
```diff
@@ -6,19 +6,24 @@ import Sentry
...
-            options.dsn = capOptions.getString("dsn") ?? ""
-            options.debug = capOptions.getBool("debug") ?? false
-            options.environment = capOptions.getString("environment") ?? "production"
-            options.enableAutoSessionTracking = capOptions.getBool("enableAutoSessionTracking") ?? false
-            options.sessionTrackingIntervalMillis = capOptions.options["sessionTrackingIntervalMillis"] as? UInt ?? 5000
-            options.attachStacktrace = capOptions.getBool("attachStacktrace") ?? false
+        let options = capOptions;
+        
+        options.beforeSend = { event in
+            if (event.exceptions.firstObject.type != nil && ["Unhandled JS Exception"].location != NSNotFound) {
```
bruno-garcia, line 12: This assumes it has the same type as it does on React Native. Might not be the same on Capacitor.

Also, does it assume there's always one exception and it has a type field?

`event.exceptions.firstObject.type ` maybe we want to check for nulls here (might be also relevant on RN /cc @jennmueng )

jennmueng: Hmm this is straight from React Native, I've never seen any issues with it though. But yeah we should do a null check just in case.

bruno-garcia: You mean no one complained about it, right? :D
Because if it fails there it means the event would be lost and the developer wouldn't know about it.

jennmueng: So over in RN the iOS native module is in objective-c and I just found out it has nil messaging, so accessing `event.exceptions.firstObject.type` shouldn't throw anything.

However here on swift you might want to use optional chaining `?.` to get the same behavior.

omercnet:
can someone merge this? plugin is currently broken in capacitor 3

jennmueng:
@omercnet We're working on a new iOS PR right now, but as we don't support the native layer for iOS yet, you can get away with it by initializing the SDK with:
```js
Sentry.init({
  dsn: "___DSN___",
  enableNative: Capacitor.getPlatform() === "android"
})
```

PR #20, opened 2020-12-23
update: get instances of user sorted

cSparkle:
User function setup for ios. No testing done yet as it will be more feasible once the entire iOS swift controller is setup.

### DIFF
--- a/ios/Plugin/Plugin.swift
+++ b/ios/Plugin/Plugin.swift
@@ -22,11 +22,27 @@ public class SentryCapacitor: CAPPlugin {
         capOptions.resolve(["value": true]);
     }
     
-    @objc public func setLogLevel() {}
-    
-    @objc private func logLevel() {}
-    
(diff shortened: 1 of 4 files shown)


ios/Plugin/Plugin.swift
```diff
@@ -22,11 +22,27 @@ public class SentryCapacitor: CAPPlugin {
...
     }
     
-    @objc public func setLogLevel() {}
-    
-    @objc private func logLevel() {}
-    
-    @objc public func setUser() {}
+    @objc public func setUser(call: CAPPluginCall) {
+        SentrySDK.configureScope { (scope) in
+            if (call.options["user"] != nil && call.options["otherUserKeys"] != nil) {
```
jennmueng, line 27: I think this should be `==` not `!=` as you're setting the user to be `nil` if both options are `nil` correct?

PR #35, opened 2021-05-24
feat: Prepare for Android alpha + refactor the SDK to be used in conjunction with @sentry/angular

jennmueng:
Refactors the SDK to be used in conjunction with `@sentry/angular`, users will need to install both `@sentry/capacitor` and `@sentry/angular` if using angular, or just `@sentry/capacitor` on its own. Currently only Android is still supported.

As this PR contains tons of removals and files, for ease of review, the most notable changes are in `wrapper.ts`, `sdk.ts`, and `SentryCapacitor.java`. 

## Notable Changes

- New init strategy as described below. The SDK is now more lightweight and just extends a sibling SDK.
- Adds support for `Sentry.nativeCrash()`.
- Catches all unexpected errors unlike before.

## Usage:
For example, when using with angular:

Install via
```bash
yarn add @sentry/capacitor @sentry/angular
```

Set up via
```typescript
// app.module.ts

import * as Sentry from "@sentry/capacitor";
import { init as sentryAngularInit, createErrorHandler }  from "@sentry/angular";

// Init by passing the sibling SDK's init as the second parameter.
Sentry.init({
  dsn: "...",
}, sentryAngularInit);

// Attach the Sentry error handler
@NgModule({
  providers: [
    {
      provide: ErrorHandler,
      useValue: createErrorHandler(),
    },
    // ...
  ],
  // ...
})
```

In theory installing alongside `@sentry/react` and `@sentry/vue` should work, but I have yet to test this so we don't officially support them yet. #40, #41

## Source Maps

When using with ionic angular and building via `ionic build`, you can upload the `www` directory with sentry-cli to upload the sourcemaps to de-minify the stack trace.
```
sentry-cli releases files "{RELEASE}" upload-sourcemaps ./www --rewrite --dist "{DIST}"
```
In the near future we should introduce auto source map uploading at bulildtime: https://github.com/getsentry/sentry-capacitor/issues/39


android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
```diff
@@ -62,14 +62,15 @@ public void startWithOptions(final PluginCall capOptions) {
         SentryAndroid.init(
             this.getContext(),
             options -> {
+                if (capOptions.getData().has("debug") && capOptions.getBoolean("debug")) {
+                    options.setDebug(true);
+                    logger.setLevel(Level.INFO);
```
bruno-garcia, line 67: We usually have `DEBUG` by default. User can lower it if it's too noisy, but out of the box we give as much information as we can

example/ionic-angular/android/app/src/main/AndroidManifest.xml
```diff
@@ -8,7 +8,9 @@
         android:label="@string/app_name"
         android:roundIcon="@mipmap/ic_launcher_round"
         android:supportsRtl="true"
-        android:theme="@style/AppTheme">
+        android:theme="@style/AppTheme"
+        android:usesCleartextTraffic="true">
```
marandaneto, line 12: its a sample so it does not matter but are we calling an HTTP endpoint instead of HTTPS?

bruno-garcia (approved):
one nit but LGTM 
:shipit:

android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
marandaneto: logger.setLevel(Level.INFO); is only within the `SentryCapacitor.java`, its not the Android SDK logger, so its fine

example/ionic-angular/src/app/app.module.ts
```diff
@@ -7,30 +7,54 @@ import { SplashScreen } from '@ionic-native/splash-screen/ngx';
...
+//     handleError(error: any) {
+//       super.handleError(error);
+//       Sentry.captureException(error.originalError ?? error);
+//     }
+//   }
+
+//   return SentryIonicErrorHandler;
+// };
+
+// const SentryIonicErrorHandler = withSentryIonicErrorHandler(ErrorHandler);
```
marandaneto, line 57: should this be enabled by default then in our sample?

android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
```diff
@@ -62,14 +62,15 @@ public void startWithOptions(final PluginCall capOptions) {
         SentryAndroid.init(
```
marandaneto, line 62: side note: bump the Android SDK to the latest GA, 5.0.0

example/ionic-angular/src/app/app.module.ts
jennmueng: This is old test code that I commented out, let me remove it

android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
jennmueng: Sounds good, after this PR I'll make one to bump Android and another to bump JS

example/ionic-angular/android/app/src/main/AndroidManifest.xml
jennmueng: Yes, I had to add this to get the capacitor dev server to work

PR #42, opened 2021-06-02
meta: Updated readme with new installation and init methods

jennmueng:
Updates the readme with the new installation for angular and the init method from #35


bruno-garcia (approved):
Discord link was a nice touch ❤️

PR #46, opened 2021-06-04
build(android): Bump sentry-android to 5.0.1

jennmueng:
Bumps the Sentry Android SDK to 5.0.1.

Additional Changes:
- Updates the `captureEnvelope` method to get the outboxPath from `HubAdapter`. A recent fix in React Native.
- Adds the native SDK's packages to the sdk info on events sent from the JS layer.
- Moves the event origin tag into its own method to align with other sdks.

Tested on the sample app with a multitude of events including native crashes.

Screen Shot 2021-06-04 at 7 41 41 PM
Screen Shot 2021-06-04 at 7 42 12 PM

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -51,7 +51,7 @@ repositories {
 dependencies {
     implementation fileTree(dir: 'libs', include: ['*.jar'])
     implementation project(':capacitor-android')
-    implementation 'io.sentry:sentry-android:5.0.0'
+    implementation 'io.sentry:sentry-android:5.0.1'
     testImplementation "junit:junit:$junitVersion"
     androidTestImplementation "androidx.test.ext:junit:$androidxJunitVersion"
     androidTestImplementation "androidx.test.espresso:espresso-core:$androidxEspressoCoreVersion"
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -52,7 +52,6 @@ dependencies {
     implementation fileTree(dir: 'libs', include: ['*.jar'])
     implementation project(':capacitor-android')
     implementation 'io.sentry:sentry-android:5.0.0'
-    implementation 'com.google.code.gson:gson:2.8.6'
     testImplementation "junit:junit:$junitVersion"
     androidTestImplementation "androidx.test.ext:junit:$androidxJunitVersion"
     androidTestImplementation "androidx.test.espresso:espresso-core:$androidxEspressoCoreVersion"
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -51,7 +51,7 @@ repositories {
 dependencies {
     implementation fileTree(dir: 'libs', include: ['*.jar'])
     implementation project(':capacitor-android')
-    implementation 'io.sentry:sentry-android:4.3.0'
+    implementation 'io.sentry:sentry-android:5.0.0'
     implementation 'com.google.code.gson:gson:2.8.6'
     testImplementation "junit:junit:$junitVersion"
     androidTestImplementation "androidx.test.ext:junit:$androidxJunitVersion"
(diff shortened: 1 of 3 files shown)


jennmueng:
Drafted to await for SentryLevel.Log filtering

android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
```diff
@@ -206,17 +189,18 @@ public void fetchRelease(PluginCall call) {
 
     @PluginMethod
     public void captureEnvelope(PluginCall call) {
-        String envelope = call.getString("envelope");
         try {
-            File installation =  new File(sentryOptions.getOutboxPath(), UUID.randomUUID().toString());
+            String envelope = call.getString("envelope");
+            final String outboxPath = HubAdapter.getInstance().getOptions().getOutboxPath();
```
marandaneto, line 194: you need to check if `outboxPath` isnt null or empty and bail out if so otherwise it'll crash

android/build.gradle
```diff
@@ -51,7 +51,7 @@ repositories {
 dependencies {
     implementation fileTree(dir: 'libs', include: ['*.jar'])
     implementation project(':capacitor-android')
-    implementation 'io.sentry:sentry-android:4.3.0'
+    implementation 'io.sentry:sentry-android:5.0.0'
```
marandaneto, line 54: similar to https://github.com/getsentry/sentry-react-native/pull/1560#discussion_r646348368

android/build.gradle
```diff
@@ -51,7 +51,7 @@ repositories {
...
     implementation 'com.google.code.gson:gson:2.8.6'
```
marandaneto, line 55: just noticed this, sentry-android already bundles `gson:gson`, why do we have its dependency here? I'd remove it and let sentry-android handle it

jennmueng: Hmm, it was used here to iterate through breadcrumb data: https://github.com/getsentry/sentry-capacitor/blob/fef9ff043778740fc8f54883f2cff05652ce3659/android/src/main/java/io/sentry/capacitor/SentryCapacitor.java#L290-L300

But we don't need it as we can just iterate through the JSONObject like we do on the other SDKs...let's refactor it

https://github.com/getsentry/sentry-cordova/blob/2d3efdc8bb0fb038b8cc3d60c2838e0e9a22973f/src/android/io/sentry/SentryCordova.java#L343-L352

android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
```diff
@@ -192,6 +192,13 @@ public void captureEnvelope(PluginCall call) {
         try {
             String envelope = call.getString("envelope");
             final String outboxPath = HubAdapter.getInstance().getOptions().getOutboxPath();
+
+            if (outboxPath != null && !outboxPath.trim().isEmpty()) {
```
marandaneto, line 196: I guess this check is wrong but I can see that Ive written wrongly as well in the comment.
so ideally:

```
if (outboxPath == null || outboxPath.isEmpty() {
  // log and return
}
```

jennmueng: Oh damn my mistake I was doing this quick while having dinner and was planning to run it after 😅

android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
```diff
@@ -206,17 +189,25 @@ public void fetchRelease(PluginCall call) {
 
     @PluginMethod
     public void captureEnvelope(PluginCall call) {
-        String envelope = call.getString("envelope");
         try {
-            File installation =  new File(sentryOptions.getOutboxPath(), UUID.randomUUID().toString());
+            String envelope = call.getString("envelope");
+            final String outboxPath = HubAdapter.getInstance().getOptions().getOutboxPath();
+
+            if (outboxPath == null && outboxPath.isEmpty()) {
```
marandaneto, line 196: this will never work, it should be `||`

jennmueng: I think I need a break 🥲

HazAT (approved):

PR #47, opened 2021-06-04
feat: Add event origin tag and also move sdkInfo into an integration

jennmueng:
Adds the `event.origin` and `event.environment` tag as a default integration. Also refactors the sdk info event processor into a default integration.

Tested on sample app with multiple events.

Screen Shot 2021-06-04 at 7 41 26 PM

### DIFF
--- a/src/integrations/eventorigin.ts
+++ b/src/integrations/eventorigin.ts
@@ -1,0 +1,28 @@
+import { EventProcessor, Integration } from '@sentry/types';
+
+/** Default EventOrigin instrumentation */
+export class EventOrigin implements Integration {
+  /**
+   * @inheritDoc
+   */
+  public static id: string = 'EventOrigin';
+
+  /**
+   * @inheritDoc
+   */
+  public name: string = EventOrigin.id;
+
+  /**
+   * @inheritDoc
+   */
+  public setupOnce(addGlobalEventProcessor: (e: EventProcessor) => void): void {
+    addGlobalEventProcessor(event => {
+      event.tags = event.tags ?? {};
+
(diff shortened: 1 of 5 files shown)


src/integrations/eventorigin.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  /**
+   * @inheritDoc
+   */
+  public setupOnce(addGlobalEventProcessor: (e: EventProcessor) => void): void {
+    addGlobalEventProcessor(event => {
+      event.tags = event.tags ?? {};
+
+      event.tags['event.origin'] = 'javascript';
+      event.tags['event.environment'] = 'javascript';
```
bruno-garcia, line 23: What other value can this be?
I'm not clear on the difference between these two tags

jennmueng: @bruno-garcia `event.origin` currently would be either `javascript` or `android`. But for `event.environment` we have that to distinguish between ndk and java like below. `event.environment` was added later and is not present on the JS layer for RN and Cordova. I added it here for consistency so that the tag is present on all events.

https://github.com/getsentry/sentry-capacitor/blob/8e50fc8159ae757e98d54534ffc539194a6b7b9b/android/src/main/java/io/sentry/capacitor/SentryCapacitor.java#L320-L340

HazAT (approved):

src/integrations/eventorigin.ts
bruno-garcia: That makes sense, thanks

bruno-garcia (approved):

PR #53, opened 2021-06-21
meta: Move back to npm in sample app due to yarn issues

jennmueng:
Yarn ignores the `files` inside the `package.json` and the `.npmignore` file when using local dependencies, this causes the whole folder to be copied into the sample app's `node_modules`, including the sample app itself.


bruno-garcia (approved):
:sadpanda:

PR #54, opened 2021-06-21
feat: Add support for Capacitor v3

### LINKED ISSUE #51: Capacitor 3
Since v3 of Capacitor many things changed and some things are different now. Would be pretty nice to support v3. We are unable to migrate to v3 because of dependency conflicts.

> https://capacitorjs.com/docs/updating/3-0

With that, it might also be a good chance to polish things up like exports:

```ts
import * as Sentry from "@sentry/capacitor";
// to
import { Sentry } from "@sentry/capacitor";
```

jennmueng:
Bumps to the latest major Capacitor v3 without needing to drop Capacitor v2.

- Adds a second sample app in `/example` (maybe rename to `/examples`) that runs v2.

Resolves #51 

See: [Capacitor v3 Upgrade Guide](https://capacitorjs.com/docs/updating/3-0) and [Capacitor v3 Plugin Upgrade Guide](https://capacitorjs.com/docs/updating/plugins/3-0)


yoyo930021:
Do we not support v2 anymore?

jennmueng:
@yoyo930021 Yes, we will not be able to support v2 going forwards by supporting v3. If you need v2 support you would have to stay on `0.1.0`

src/definitions.ts
```diff
@@ -2,18 +2,18 @@ import { Breadcrumb, Response } from '@sentry/types';
 
 import { CapacitorOptions } from './options';
 
-declare module '@capacitor/core' {
-  interface PluginRegistry {
-    SentryCapacitor: SentryCapacitorPlugin;
-  }
-}
+// declare module '@capacitor/core' {
```
bruno-garcia, line 5: This will get merged?

bruno-garcia (approved):

src/definitions.ts
jennmueng: Left it there because I was not sure if we were going to fully drop v2 or not, turns out we don't need it regardless because we're not exposing the native bridge to the users. This just types the native bridge plugin.

jennmueng:
@yoyo930021 Disregard my last comment. We will continue supporting v2 and will drop it when v4 hits.

.gitignore
```diff
@@ -18,6 +18,17 @@ example/ionic-angular/www
 example/ionic-angular/.gradle
 example/ionic-angular/build
 example/ionic-angular/.ionic
+example/ionic-angular/.yalc
+
+example/ionic-angular-v2/dist
```
bruno-garcia, line 23: Would this work to include both samples?

```suggestion
example/ionic-angular*/dist
```

example/ionic-angular-v2/src/app/app.module.ts
```diff
@@ -0,0 +1,56 @@
...
+import { AppComponent } from './app.component';
+
+// ATTENTION: Change the DSN below with your own to see the events in Sentry. Get one at sentry.io
+Sentry.init(
+  {
+    dsn:
+      'https://4079af8b316240ea9453eb0a23b715cc@o447951.ingest.sentry.io/5522756',
+    // An array of strings or regexps that'll be used to ignore specific errors based on their type/message
+    ignoreErrors: [/MiddleEarth_\d\d/, 'RangeError'],
+    // Debug mode with valuable initialization/lifecycle information
```
bruno-garcia, line 23: What about something like this?
```suggestion
    // To see what the Sentry SDK is doing; Helps when setting things up
```

example/ionic-angular-v2/src/app/app.module.ts
```diff
@@ -0,0 +1,56 @@
...
+    debug: true,
+    // Whether SDK should be enabled or not
+    enabled: true,
```
bruno-garcia, line 26: Isn't this `true` by default? Generally we have that inferred by having a DSN or not (or `""` which means disable SDK)

bruno-garcia (approved):
some nits but LGTM

example/ionic-angular-v2/src/app/app.module.ts
jennmueng: Yes it is, this was copied from the original sample, I think it was added just to show the option?

PR #57, opened 2021-06-28
feat: Add support for iOS

jennmueng:
Adds support for the native layer on iOS.

## Notes:
- No scope sync yet.
- Added support for session tracking
- Added event origin and environment tags.
- Fetches the SdkInfo from `sentry-cocoa` across the bridge to JS and adds it to all events' `sdk.packages` (Due to `sentry-cocoa` not calling `beforeSend` on `captureEnvelope`
- Works for both apps running Capacitor 2 and 3. However, apps on Capacitor 2 will have to bump the minimum iOS deployment target to `12.0` due to us requiring it to use with Capacitor 3.


HazAT (approved):
Skimmed over it, lgtm

example/ionic-angular-v2/android/app/src/main/assets/capacitor.config.json
```diff
@@ -2,6 +2,9 @@
   "appId": "io.ionic.starter",
   "appName": "ionic-angular",
   "bundledWebRuntime": false,
+  "ios": {
+    "minVersion": "12.0"
```
bruno-garcia, line 6: Is this an SDK imposed limit? If so, we should note this on the docs

example/ionic-angular/package.json
```diff
@@ -8,6 +8,7 @@
     "@angular/platform-browser-dynamic": "~10.0.0",
     "@angular/router": "~10.0.0",
     "@capacitor/core": "3.0.1",
+    "@capacitor/ios": "^3.0.2",
```
bruno-garcia, line 11: Does it need to match the above?

src/integrations/sdkinfo.ts
```diff
@@ -14,17 +15,30 @@ export class SdkInfo implements Integration {
...
    * @inheritDoc
    */
   public setupOnce(addGlobalEventProcessor: (e: EventProcessor) => void): void {
-    addGlobalEventProcessor(event => {
+    addGlobalEventProcessor(async event => {
+      // The native SDK info package here is only used on iOS as `beforeSend` is not called on `captureEnvelope`.
+      // this._nativeSdkInfo should be defined a following time so this call won't always be awaited.
+      if (NATIVE.platform === 'ios' && this._nativeSdkInfo === null) {
+        try {
+          this._nativeSdkInfo = await NATIVE.fetchNativeSdkInfo();
```
bruno-garcia, line 29: This value never changes at runtime so we could cache this and return the same value and avoid the bridge call

src/integrations/sdkinfo.ts
```diff
@@ -14,17 +15,30 @@ export class SdkInfo implements Integration {
...
+        } catch (_) {
+          // Do nothing if this fails, we would rather have the event be sent with a package missing.
```
bruno-garcia, line 31: We should log this out so we can debug it?

bruno-garcia (approved):
Some suggestions but generally ready to ship!
Congrats!

example/ionic-angular-v2/android/app/src/main/assets/capacitor.config.json
jennmueng: Nope, this is a Capacitor 3 requirement. It still works with capacitor 2 but they'll either need to bump their version up to 12 like in this config json, or they can write a script in their pod file to force it back down.

example/ionic-angular/package.json
jennmueng: Actually, yes it should

src/integrations/sdkinfo.ts
jennmueng: Hmm I don't see any way of caching this other than modifying the JS at build time right? Storing it as a file outside JS would also need an async bridge call. This bridge call should only be called once so it should be fine IMO

PR #61, opened 2021-07-05
feat: Finish scope sync for both iOS and Android

jennmueng:
Add scope sync methods for `setTag`, `setExtra`, `setUser`, `addBreadcrumb`, `clearBreadcrumb` for iOS. 

Fixes `setUser` not taking non-default data and `clearBreadcrumbs` on Android.

Adds test for the JS side of the bridge, and methods to test scope sync on the sample apps.

Tested on the v3 sample app.


HazAT (approved):
same, skimmed over it, lgtm

jennmueng:
@philipphofmann Do you mind having a quick run through of this as it contains a lot of swift? (and the fact that I haven't really done much swift before) Thanks!

PR #111, opened 2022-02-02
feat(deps): Bump Sentry SDKs to 6.17.4

AbhiPrasad:
- No longer export deprecated Status enum
- Remove guard from transport `sendEvent` as it is redundant


bruno-garcia (approved):
Thanks Abhi!

PR #126, opened 2022-02-17
fix: Update logic for RewriteFrames

kamilogorek:
`app:///` is only required for custom protocols, where any valid host, eg. `http(s)://example.com` is matched by a default cli `~/` prefix, which acts as a host wildcard.
Keeping `http(s)` hosts intact, also allows us to fetch sources from the internet if that's configured (not even sure if that's doable in capacitor, but I assume yes?).


bruno-garcia (approved):
Nice. Thanks

PR #139, opened 2022-02-25
remove sdkProcessingMetadata from native events

### LINKED ISSUE #125: "Discarded unknown attribute" problem flagged on events
### Overview

Events in Sentry have a problem flagged that says `Discarded unknown attribute`.

On iOS the problem will be
- `sdkProcessingMetadata: Discarded unknown attribute`

On Android the problem will be
- `unknown: Discarded unknown attribute`

This occurs on hybrid iOS and hybrid Android which use the native SDKs to process and send the event (might be relevant). It is not an issue on Web wherein it works fine (it also works fine using `@sentry/browser` directly).

A repro can be found [here](https://github.com/KevinKelchen/sentry-capacitor-unknown-metadata-issue).

### Environment

How do you use Sentry?
Sentry SaaS (sentry.io)

Which SDK and version?
Sentry Capacitor - 0.4.2

### Steps to Reproduce

See [Steps to Reproduce](https://github.com/KevinKelchen/sentry-capacitor-unknown-metadata-issue#steps-to-reproduce).

### Expected Result

The event from hybrid iOS and Android appears in Sentry without a problem being flagged.

We normally do not have problems flagged on our events from production builds. If the red problem banner were to be accepted as always appearing due to this issue, I think people would initially think something is really wrong. After being told to ignore it due to this issue, I think they then would start to ignore the problem banner altogether which would make it less likely for people to notice different, more actionable problems.

### Actual Result

The event from hybrid iOS and Android appears in Sentry with a problem being flagged.

### Final Thoughts

Thank you so much for taking the time to investigate and/or work on this issue! 😀

lucas-zimerman:
This parameter is only used on the Javascript side of the SDK and shouldn't be sent to the Native event handler.

Closes #125.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- remove sdkProcessingMetadata from native events ([#139](https://github.com/getsentry/sentry-capacitor/pull/139))
+
 ## 0.4.3
 
 ### Fixes
--- a/src/wrapper.ts
+++ b/src/wrapper.ts
@@ -24,6 +24,9 @@ export const NATIVE = {
 
     const event = this._processLevels(_event);
 
+    // Delete this metadata as this should not be sent to Sentry.
+    delete event.sdkProcessingMetadata;
+
     const header = {
       event_id: event.event_id,
       sdk: event.sdk,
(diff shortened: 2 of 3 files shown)


jennmueng (approved):
LGTM

PR #145, opened 2022-03-04
disable native sdk when enablenative is false

### LINKED ISSUE #137: Disabling the SDK still results in native errors being sent
### Overview

Disabling the SDK still results in native errors being sent.

There are 2 scenarios:
- Scenario 1 is that `enabled` and `enableNative` flags at initialization do not stop Android native errors from being sent.
- Scenario 2 is that disabling the SDK at runtime does not stop Android and iOS native errors from being sent.

This occurs on hybrid iOS and hybrid Android which use the native SDKs to process and send the event (might be relevant). It is not an issue on Web wherein it works fine (it also works fine using `@sentry/browser` directly).

A repro can be found [here](https://github.com/KevinKelchen/sentry-capacitor-enable-sdk-issue).

### Environment

How do you use Sentry?
Sentry SaaS (sentry.io)

Which SDK and version?
Sentry Capacitor - 0.4.3

### Steps to Reproduce

See [Steps to Reproduce](https://github.com/KevinKelchen/sentry-capacitor-enable-sdk-issue#steps-to-reproduce).

### Expected Result

- Scenario 1
  - Expected setting `enabled: false` (and perhaps even `enableNative: false`) to be sufficient at stopping native errors from being sent on Android. `enabled: false` alone works as expected on iOS. `enabled: false` alone works as expected for Android *JavaScript* errors.
- Scenario 2
  - Expected setting `getCurrentHub().getClient().getOptions().enabled = false` (and perhaps even `(getCurrentHub().getClient().getOptions() as any).enableNative` with `(getCurrentHub().getClient().getOptions() as any).enableNativeCrashHandling)` at runtime to be sufficient at stopping native errors from being sent on Android and iOS.
  - Related: If we are to use `getCurrentHub().getClient().getOptions()` to disable the SDK at runtime, it would be nice if it had TypeScript typings that reflects the changes in the Capacitor SDK such as the additional native-related properties which do appear to be present at runtime.

### Actual Result

- Scenario 1
  - Setting `enabled: false` (even with `enableNative: false`) results in native errors being sent on Android.
- Scenario 2
  - Setting `getCurrentHub().getClient().getOptions().enabled = false` (even with `(getCurrentHub().getClient().getOptions() as any).enableNative` and `(getCurrentHub().getClient().getOptions() as any).enableNativeCrashHandling)` at runtime results in native errors being sent on Android and iOS.
  - Related: Using `getCurrentHub().getClient().getOptions()` to disable the SDK at runtime does not have TypeScript typings that reflects the changes in the Capacitor SDK such as the additional native-related properties which do appear to be present at runtime.

The TypeScript typings enhancement could be its own issue, but I erred on the side of including it in here for now in case there's a preference to keep all of this SDK enable/disable stuff in one place. Also, if there's a better approach for disabling the SDK at runtime that results in not using the untyped native-related properties of those APIs, then our use case for them may go away. 🙂

### Final Thoughts

Thank you so much for taking the time to investigate and/or work on this issue! 😀

lucas-zimerman:
Close #137

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@
 
 ### Fixes
 
-- remove sdkProcessingMetadata from native events ([#139](https://github.com/getsentry/sentry-capacitor/pull/139))
+- Disable native sdk when enabled/enablenative is set to false ([#145](https://github.com/getsentry/sentry-capacitor/pull/145))
+- Remove sdkProcessingMetadata from native events ([#139](https://github.com/getsentry/sentry-capacitor/pull/139))
 
 ## 0.4.3
 
--- a/src/sdk.ts
+++ b/src/sdk.ts
@@ -31,11 +31,9 @@ export function init<O>(
     finalOptions.enableNative = false;
     finalOptions.enableNativeNagger = false;
   } else {
-    finalOptions.enableNativeNagger = true;
-
-    if (finalOptions.enableNative == null) {
-      finalOptions.enableNative = true;
-    }
+    // keep the original value if user defined it.
+    finalOptions.enableNativeNagger ??= true;
+    finalOptions.enableNative ??= true;
   }
 
   const capacitorHub = new Hub(undefined, new CapacitorScope());
--- a/src/sdk.ts
+++ b/src/sdk.ts
@@ -26,12 +26,16 @@ export function init<O>(
     ...options,
   };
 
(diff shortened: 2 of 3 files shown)


bruno-garcia (approved):

bruno-garcia:
Thank you Lucas

Issue #127, opened 2022-02-17
Native Errors: beforeSend callback not executed before sending event to server


BertoDBQ:
### Overview

Native IOS errors (ie OutOfMemory) are not going thru the beforeSend callback event before being sent to the server.

Would like to use the beforeSend callback to filter out native errors and not log them to Sentry in development mode.  

Have been able to get around this by setting 'enableNative' equal to false.  Is there some equivalent to beforeSend for native errors that may be used?  (maybe this is a feature request vs bug???)    

### Environment

How do you use Sentry?
Sentry SaaS (sentry.io) 

Which SDK and version?
Sentry Capacitor - 0.4.2

### Expected Result

Expected native errors to flow thru the beforeSend callback before being sent to Sentry.

lucas-zimerman:
Native events can only be called from the Native before send and not from the JavaScript callback so you'll need to check the iOS layer to filter the native layers.

This is not a trivial change and won't be altered on Sentry Capacitor.

PR #147, opened 2022-03-08
Add Missing config `enableOutOfMemoryTracking` for iOS/Mac

lucas-zimerman:
Add an option filter to avoid registering out of error events if the user doesn't want it.

Partially solves #127

### DIFF
--- a/src/options.ts
+++ b/src/options.ts
@@ -35,4 +35,14 @@ export interface CapacitorOptions
 
   /** When enabled, all the threads are automatically attached to all logged events on Android */
   attachThreads?: boolean;
+
+  /**
+  * Enables Out of Memory Tracking for iOS and macCatalyst.
+  * See the following link for more information and possible restrictions:
+  * https://docs.sentry.io/platforms/apple/guides/ios/configuration/out-of-memory/
+  *
+  * @default true
+  * */
+  enableOutOfMemoryTracking?: boolean;
+
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixes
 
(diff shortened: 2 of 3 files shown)


src/options.ts
```diff
@@ -35,4 +35,14 @@ export interface CapacitorOptions
...
   attachThreads?: boolean;
+
+  /**
+  * Enables Out of Memory Tracking for iOS and macCatalyst.
+  * See the following link for more information and possible restrictions:
+  * https://docs.sentry.io/platforms/apple/guides/ios/configuration/out-of-memory/
+  *
+  * @default true
+  * */
+  enableOutOfMemoryTracking?: boolean;
```
bruno-garcia, line 46: Do we have this on the React Native SDK? I wonder if this is a sign we should have this TypeScript type definition live elsewhere and use it in both capacitor+RN
cc @marandaneto @jennmueng

lucas-zimerman: Yes, it's also on React Native
https://github.com/getsentry/sentry-react-native/blob/3f0c1707d86c1b6e28582d1db21926f2ac543f0b/src/js/options.ts#L76-L83

marandaneto: a Unified package would not be bad but each platform has its specific options, not sure how much we'd gain with it.
Not a bad idea tho.

piotr-cz:
This option is missing in [Capacitor docs](https://docs.sentry.io/platforms/javascript/guides/capacitor/configuration/options/#hybrid-sdk-options) (available in [React Native docs](https://docs.sentry.io/platforms/react-native/configuration/options/#enableOutOfMemoryTracking))

lucas-zimerman:
Thanks for pointing about the Documentation.
You can track the changes here: https://github.com/getsentry/sentry-docs/pull/5118

PR #148, opened 2022-03-09
Refactor docs

lucas-zimerman:
- Add links to contributing page.

#skip-changelog.

reduced the goal of this PR to  just to add the links to the contributing page.

### DIFF
--- a/example/ionic-angular/README.md
+++ b/example/ionic-angular/README.md
@@ -33,17 +33,3 @@ Then on the ionic-angular folder
 ### Web
 
 `http-server ./www`
-
-# Troubleshooting
-
-## Mac M1
-
-### Error on step `Sync resources`
-
-You may face the following error during `npx cap sync`
-
-```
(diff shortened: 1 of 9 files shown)


example/ionic-angular/README.md
```diff
@@ -33,3 +33,17 @@ Then on the ionic-angular folder
 ### Web
 
 `http-server ./www`
+
+# Troubleshooting
```
bruno-garcia, line 37: This is for ppl using the sdk or developing? If it's for users it should be on the docs Instead

lucas-zimerman: It could happen with anyone trying a capacitor project on a Mac M1.

It's not a problem with the SDK but more like a guide to folks trying the examples on an ARM device.

bruno-garcia: I think that should be pointed out. It can be confusing: Troubleshooting if you're using the SDK?

bruno-garcia (approved):

Issue #90, opened 2021-10-05
Source maps are not linked correctly by Sentry


StErMi:
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io) 

Which SDK and version?

```
    "@sentry/capacitor": "^0.4.0",
    "@sentry/react": "^6.11.0",
    "@sentry/tracing": "^6.11.0",
```

### Description

Source maps are not linked correctly by Sentry and I'm unable to understand how to fix the code.
This is the error trace

```
RangeError: Invalid time value
  at Z(app:///static/js/2.8a9aba35.chunk.js:2:718874)
  at un(app:///static/js/main.b1e0cf1d.chunk.js:1:70866)
  at ? (app:///static/js/main.b1e0cf1d.chunk.js:1:71066)
  at ? (app:///static/js/main.b1e0cf1d.chunk.js:1:71975)
  at ri(app:///static/js/main.b1e0cf1d.chunk.js:1:87954)
  at ou(app:///static/js/2.8a9aba35.chunk.js:2:786602)
  at Ga(app:///static/js/2.8a9aba35.chunk.js:2:839060)
  at ws(app:///static/js/2.8a9aba35.chunk.js:2:826234)
  at Ss(app:///static/js/2.8a9aba35.chunk.js:2:826162)
  at Cs(app:///static/js/2.8a9aba35.chunk.js:2:826025)
  at ms(app:///static/js/2.8a9aba35.chunk.js:2:822991)
  at ? (app:///static/js/2.8a9aba35.chunk.js:2:772370)
  at t.unstable_runWithPriority(app:///static/js/2.8a9aba35.chunk.js:2:849789)
  at zo(app:///static/js/2.8a9aba35.chunk.js:2:772147)
  at Yo(app:///static/js/2.8a9aba35.chunk.js:2:772315)
  at Wo(app:///static/js/2.8a9aba35.chunk.js:2:772250)
  at ds(app:///static/js/2.8a9aba35.chunk.js:2:820360)
  at Tu(app:///static/js/2.8a9aba35.chunk.js:2:791927)
  at ? (app:///static/js/2.8a9aba35.chunk.js:2:508500)
  at ? (app:///static/js/2.8a9aba35.chunk.js:2:211534)
  at e.notifyFn(app:///static/js/2.8a9aba35.chunk.js:2:211092)
  at ? (app:///static/js/2.8a9aba35.chunk.js:2:211699)
  at Array.forEach(app:///<anonymous>)
[...]

PR #162, opened 2022-04-22
Fix Stack Trace when using localhost

lucas-zimerman:
This PR fixes the following cases found on #90

* localhost with port number was incorrectly parsed. (http://localhost:8080/)
* localhost addresses became app:/// ... not allowing the correct source map to be applied (http://localhost/test.js became app:///test.js instead of /test.js)

Additionally, extra tests were added to the StackTrace parser.

With the following changes, users will be able to get the correct source map when accessing their apps on the browser from  localhost with and without a port, but also by specifying the local ip address from the host machine.

### DIFF
--- a/test/sdk.test.ts
+++ b/test/sdk.test.ts
@@ -1,4 +1,6 @@
-import { BrowserOptions } from '@sentry/browser';
+import { BrowserOptions, StackFrame } from '@sentry/browser';
+import { RewriteFrames } from '@sentry/integrations';
+import { Integration } from '@sentry/types';
 
 import { CapacitorOptions } from '../src';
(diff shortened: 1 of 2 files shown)


test/sdk.test.ts
```diff
@@ -29,67 +31,67 @@ describe('SDK Init', () => {
...
-      'ios',
-      { dsn: '', enableAutoSessionTracking: false },
-      false,
-      {
-        enableNative: true,
-        enableAutoSessionTracking: false,
-      },
-    ],
-  ];
+      [
```
lucas-zimerman, line 34: CI decided to format all those lines

marandaneto (approved):

PR #163, opened 2022-04-25
fix update yarn.lock

lucas-zimerman:
Release.yml doesn't like when it produces uncommitted changes to the local branch.

Ideally in the future this check should be done during the PR and not on the release phase.

#skip-changelog.


bruno-garcia (approved):

bruno-garcia:
Bumping deps should always include committing these

bruno-garcia:
Thanks for fixing it

Issue #176, opened 2022-05-26
iOS warnings in Xcode 13.4


piotr-cz:
### Environment

How do you use Sentry?

- Sentry SaaS (sentry.io)

Which SDK and version?

- @sentry/capacitor 0.6.1
- @sentry/react 6.19.2

### Steps to Reproduce

1. Build project from Xcode (Product > Run)
2. Issues related to SentryCapacitor appear in Issue Navigator

### Expected Result

No issues related to SentryCapacitor

### Actual Result

- Swift Compiled Warning
  - *Expression implicitly coerced from 'Int?' to 'Any'*
    Source: https://github.com/getsentry/sentry-capacitor/blob/0.6.1/ios/Plugin/Plugin.swift#L95
    - Provide a default value to avoid this warning
    - Force-unwrap the value to avoid this warning
    - Explicitly cast to 'Any' with 'as Any' to silence this warning

  - *'PluginResultData' is deprecated: renamed to 'PluginCallResultData'*
    Source: https://github.com/getsentry/sentry-capacitor/blob/0.6.1/ios/Plugin/Plugin.swift#L158
    Context: Interface has apparently been renamed [in Capacitor 3](https://github.com/ionic-team/capacitor/blob/3.5.1/ios/Capacitor/Capacitor/CAPPluginCall.swift#L3)

PR #177, opened 2022-05-26
iOS: Replace deprecated PluginResultData

piotr-cz:
PluginResultData has deprecated in favor of PluginCallResultData in Capacitor 3 (see https://github.com/ionic-team/capacitor/blob/3.5.1/ios/Capacitor/Capacitor/CAPPluginCall.swift#L3)

This package requires @capacitor/core@3 since it's 0.2.1 version

Fixes one issue of https://github.com/getsentry/sentry-capacitor/issues/176

### DIFF
--- a/ios/Plugin/Plugin.swift
+++ b/ios/Plugin/Plugin.swift
@@ -155,7 +155,7 @@ public class SentryCapacitor: CAPPlugin {
                 }
                 print("Contexts: \(debugContext ?? "")")
             }
-            call.resolve(contexts as PluginResultData)
+            call.resolve(contexts as PluginCallResultData)
         }
     }
 


lucas-zimerman (approved):
Thanks for the contribution!

aparajita:
Hey @lucas-zimerman, I see this was merged, but there was a regression:

https://github.com/getsentry/sentry-capacitor/blob/6f4f717abcd6e1e593fd058d9a0e911b0b8524ef/ios/Plugin/Plugin.swift#L175

Happened with [this commit](https://github.com/getsentry/sentry-capacitor/commit/7a36a739866a94cad6772552adc625d4f7c51e39).

Would be nice to have that fixed.

PR #178, opened 2022-05-26
iOS: Avoid warning by force-unwrapping value

piotr-cz:
Fixes first issue of https://github.com/getsentry/sentry-capacitor/issues/176.

I have no experience with swift, but since the `payloadSize` is being tested against `nil`, it should be safe for force Int

#skip-changelog.


lucas-zimerman (approved):
Thanks for the contribution!

Issue #39, opened 2021-06-02
Introduce auto source map uploading on build time


jennmueng:
Most likely we would integrate the existing Sentry Webpack plugin

muuvmuuv:
This would be spectacular! Just started using Sentry and was about to add a post build script. Since we are at an early stage, I can wait for this automation. Big love for your work into this!

lucas-zimerman:
When implementing this feature, it should also address this issue: https://github.com/getsentry/team-mobile/issues/3

muuvmuuv:
So we used the Webpack Plugin till now and it came to some problems with it after the build since the content hash on scripts differ between web and native build (two different pipelines on different providers...). I guess this would solve this because it would upload sourcemaps specifically for the native build. At the end it would mean three sourcemaps beeing uploaded; android, ios and web. Tagging these with some flag could help identify the build later in Sentry and match with its uploaded sourcemap.

PR #180, opened 2022-05-30
fix: remove unnecessary dependencies to promises and @sentry/wizard

olivier-blanc-openit:
Sentry/capacitor had a runtime dependency on Sentry/wizard, which then depends on `rxjs` through `inquirer`.

This cause issues in monorepo configurations, for instance : https://github.com/ReactiveX/rxjs/issues/6628 and https://github.com/webpack/webpack/issues/14405.

Since Sentry/wizard is more or less a CLI tool, it is not required as a runtime dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,9 +59,7 @@
     "@sentry/integrations": "6.19.2",
     "@sentry/tracing": "6.19.2",
     "@sentry/types": "6.19.2",
-    "@sentry/utils": "6.19.2",
-    "@sentry/wizard": "^1.1.4",
-    "promise": "^8.1.0"
+    "@sentry/utils": "6.19.2"
   },
   "devDependencies": {
     "@capacitor/android": "^3.0.1",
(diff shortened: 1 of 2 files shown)


lucas-zimerman:
wizard may be required once https://github.com/getsentry/sentry-capacitor/issues/39 is implemented.

bruno-garcia:
@olivier-blanc-openit could u please add a changelog entry?

bruno-garcia:
Also, I believe the wizard should become a dev dependency then, no?

lucas-zimerman:
An update: Sentry Wizard will be supported by Capacitor, the promises package will be removed from the dependencies and Sentry Wizard will be moved to the devDependencies. All that after the integration with the Wizard is completed

marandaneto:
> An update: Sentry Wizard will be supported by Capacitor, the promises package will be removed from the dependencies and Sentry Wizard will be moved to the devDependencies. All that after the integration with the Wizard is completed

Should we reopen this issue? The stale bot closed it.

lucas-zimerman:
This PR will move to https://github.com/getsentry/sentry-capacitor/pull/400
Thank you @olivier-blanc-openit  for the contribution!

PR #213, opened 2022-09-04
Feat: check missing changes 

lucas-zimerman:
The goal is to check if there aren't any missing changes that weren't pushed by the developer, if so, a release could get stalled because there are uncommitted changes on the main branch.

#skip-changelog


bruno-garcia:
I don't follow. What uncommitted code would show up there?

lucas-zimerman:
> I don't follow. What uncommitted code would show up there?

Here is an example of the runner 

https://github.com/getsentry/sentry-capacitor/runs/8188451974?check_suite_focus=true

image

It shows there are code that got modified that wasn't commited. An example is someone altering the `packages.json` and not doing `yarn install` to update the `yarn.lock` file.

The main goal from this PR is to avoid the uncommited errors on the release of new versions like this: https://github.com/getsentry/sentry-capacitor/runs/7695221326?check_suite_focus=true , with that, we'll check for such changes during a PR, avoiding the block of releasing a new version due to uncommited changes.

bruno-garcia (changes requested):
I'd defer this to @AbhiPrasad.
How do we do this in the js repo? I imagine this is standard in any repo with a lock file

AbhiPrasad:
I would install using the frozen lock file command: https://classic.yarnpkg.com/en/docs/cli/install/#toc-yarn-install-frozen-lockfile

this will throw an error on dep install if the lock file is out of sync

lucas-zimerman:
@AbhiPrasad With the given yarn command it resulted on the following error when yarn.lock is out of sync: 
image

AbhiPrasad:
This means there’s some sync issues between local env and CI

lucas-zimerman:
> This means there’s some sync issues between local env and CI

I mean, it's working as intended (the CI test 😅)

AbhiPrasad:
Ah awesome - sorry for misunderstanding just took a quick skim at this 😅

AbhiPrasad (approved):

bruno-garcia (approved):

PR #230, opened 2022-10-10
Bump Android SDK to 6.4.3

lucas-zimerman:
This PR Bumps Sentry Android 5.7.0 to 6.4.3 and also updates the contributing page.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## Unreleased
+
+### Dependencies
+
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## Unreleased
+
+### Dependencies
+
+- Bump Sentry Android SDK to `6.4.3` ([#230](https://github.com/getsentry/sentry-capacitor/pull/230))
```
kahest, line 7: I think it would be better to keep referring to the dependency as sentry-java, so as not to confuse ppl who are not aware that this is the Android SDK

kahest (approved):
LGTM!
thanks for the PR!

PR #231, opened 2022-10-10
Bump iOS SDK to 7.27.1

lucas-zimerman:
Also (without impact to the end user)
- Ignored some extra files generated by Ionic/VsCode on the Sample projects
- Fixed Podfiles that wasn't automatically updating due to the wrong path on the node_modules reference.

### DIFF
--- a/ios/Podfile
+++ b/ios/Podfile
@@ -1,4 +1,4 @@
-require_relative '../../node_modules/@capacitor/ios/scripts/pods_helpers'
+require_relative '../node_modules/@capacitor/ios/scripts/pods_helpers'
 
 platform :ios, '13.0'
 
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -1,15 +1,15 @@
 PODS:
-  - Capacitor (4.1.0):
+  - Capacitor (4.0.1):
     - CapacitorCordova
-  - CapacitorCordova (4.1.0)
-  - Sentry (7.23.0):
-    - Sentry/Core (= 7.23.0)
-  - Sentry/Core (7.23.0)
+  - CapacitorCordova (4.0.1)
+  - Sentry (7.27.1):
+    - Sentry/Core (= 7.27.1)
+  - Sentry/Core (7.27.1)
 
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## Unreleased
+
+### Dependencies
+
+- Bump Sentry iOS SDK to `2.27.1` ([#231](https://github.com/getsentry/sentry-capacitor/pull/231))
```
kahest, line 7: typo: should be `7.27.1`

kahest: also I think it would be better to keep referring to the dependency as `sentry-cocoa`, so as not to confuse ppl who are not aware that this is the iOS SDK

kahest (changes requested):
just a typo, otherwise lgtm
thanks for the PR!

ios/Podfile
```diff
@@ -12,7 +12,7 @@ end
 target 'Plugin' do
   capacitor_pods
 
-  pod 'Sentry', '7.23.0'
+  pod 'Sentry', '7.27.1'
```
kahest, line 15: the corresponding changes in Podfile.lock are missing from the PR

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@
 
 ### Dependencies
 
+- Bump Sentry iOS SDK to `2.27.1` ([#231](https://github.com/getsentry/sentry-capacitor/pull/231))
```
lucas-zimerman, line 7:
```suggestion
- Bump Sentry iOS SDK to `7.27.1` ([#231](https://github.com/getsentry/sentry-capacitor/pull/231))
```

ios/Podfile
lucas-zimerman: Indeed, thanks for pointing out! I have fixed on the same PR the `require_relative` on the podfiles. With the wrong path the podfiles.lock wasn't updated.

kahest (approved):
LGTM!

PR #245, opened 2022-10-16
ref(ci): fix set-output / set-state deprecation

lucas-zimerman:
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

#skip-changelog


kahest (approved):
LGTM! Thanks :)

PR #246, opened 2022-10-17
Update README.md

lucas-zimerman:
Sentry Capacitor requires the sibling SDKs to be exactly the same. By add the flag `--exact` to yarn, it forces it to use a specific version (in this case the latest one) instead of allowing it to install a different but similar version of Sentry Javascript/Angular/...

#skip-changelog


kahest (approved):
LGTM, Thanks!

Issue #218, opened 2022-09-22
Type 'BrowserTracing' is not assignable to type 'Integration' :: Vue 3 w/ Capactor 3


vanpyrzericj:
### Environment

```
"@capacitor/core": "^3.5.1",
"@sentry/capacitor": "^0.10.0",
"@sentry/vue": "^7.13.0,
"vue": "^3.1.5"
```

```
import * as Sentry from "@sentry/capacitor";
import * as SentryVue from "@sentry/vue";
import { BrowserTracing } from "@sentry/tracing";

Sentry.init({
        app,
        dsn: "***",
        integrations: [
            new BrowserTracing({
                routingInstrumentation: SentryVue.vueRouterInstrumentation(router),
                tracingOrigins: ["localhost", "***", "***", /^\//],
            }),
        ],
        environment: process.env.VUE_APP_SH_ENV,
        tracesSampleRate: 1.0,
    }, SentryVue.init);
```

### Steps to Reproduce

1. Use above package versions & set up in `main.ts` for a Vue3 project
2. Attempt to build project

### Expected Result

The project would build successfully.

### Actual Result

See #197 for my original comment. I took NilsFrkl's advice and removed the direct reference to `@sentry/tracing` which still produces the error:

```
TS2322: Type 'BrowserTracing' is not assignable to type 'Integration'.   
Types of property 'setupOnce' are incompatible.     
Type '(_: (callback: EventProcessor) => void, getCurrentHub: () => Hub) => void' is not assignable to type '(addGlobalEventProcessor: (callback: EventProcessor) => void, getCurrentHub: () => Hub) => void'.       
Types of parameters '_' and 'addGlobalEventProcessor' are incompatible.         
Types of parameters 'callback' and 'callback' are incompatible.          
Type 'import("/Users/vanpyrzericj/Development/abcd/app-web/node_modules/@sentry/types/types/eventprocessor").EventProcessor' is not assignable to type 'import("/Users/vanpyrzericj/Development/abcd/app-web/node_modules/@sentry/vue/node_modules/@sentry/types/types/eventprocessor").EventProcessor'.             
Types of parameters 'event' and 'event' are incompatible.               
Type 'import("/Users/vanpyrzericj/Development/abcd/app-web/node_modules/@sentry/vue/node_modules/@sentry/types/types/event").Event' is not assignable to type 'import("/Users/vanpyrzericj/Development/abcd/app-web/node_modules/@sentry/types/types/event").Event'.                 
[...]

PR #248, opened 2022-10-20
Add sibling check

### LINKED ISSUE #45: Make clear in installation docs that the sibling SDK version has to match.
We will need to document in the installation steps to ensure the version of the sibling SDK matches the version of the sentry-javascript dependencies that @sentry/capacitor depends on.

lucas-zimerman:
The goal is to better inform the user that Sentry/Capacitor requires the exact version on the Sibling SDKs as defined on the peerDependency.

This code will look for any @sentry/package references on the packages.json file from the user's project, if those references doesn't match the required one by the script, the script will warn the user about it when a dependency gets installed or updated.

Some questions in regard to this feature:

- Should it feature just warn the user or throw an error when installing/updating the dependency?

If we throw an error, the user will only be able to update the project if he/she updates all sibling dependencies in one command (ie yarn add @sentry/angular@7.15.0 @sentry/hub@7.15.0 @sentry/capacitor@0.11.0 --exact)

Fixes #45, #218, #268

### DIFF
--- a/scripts/check-siblings.js
+++ b/scripts/check-siblings.js
@@ -9,15 +9,15 @@ const updateArgument = '--update-sentry-capacitor';
  * @return {Boolean} true if requested to skip the post-install check, false otherwise.
  */
 function SkipPostInstall() {
-      if (env.npm_config_update_sentry_capacitor) {
-            //NPM.
-            return true;
-      }
-      else if (env.npm_config_argv && env.npm_config_argv.includes(updateArgument)) {
-            //YARN.
-            return true;
-      }
-      return false;
+  if (env.npm_config_update_sentry_capacitor) {
+    //NPM.
+    return true;
+  }
+  else if (env.npm_config_argv && env.npm_config_argv.includes(updateArgument)) {
+    //YARN.
+    return true;
+  }
+  return false;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


scripts/check-siblings.js
```diff
@@ -0,0 +1,36 @@
+const fs = require('fs');
+const path = require('path');
+
+//Filters all Sentry packages but Capacitor itself.
+const jsonFilter = /\s*\"\@sentry\/(?!capacitor)(?<packageName>[a-zA-Z]+)\"\:\s*\"(?<version>.+)\"/;
+const siblingVersion = "7.15.0";
```
AbhiPrasad, line 6: Can we just grab this from the sentry capacitor package.json?

lucas-zimerman: Yes we can

lucas-zimerman: Another option is to grab it from the environment inside of the postinstall script:
```
console.log("Core version is : "+ env.npm_package_dependencies__sentry_browser);
output
Core version is : 7.15.0
```
Not sure about the compatibility about it with npm or older yarn versions.

lucas-zimerman:
Added more code after validating both yarn/npm.
When using yarn, we can validate the packages in two places
1) The console arguments (`yarn install @sentry/angular@7.13.0`
In this case we can simply validate if the sentry version requested is compatible with the current capacitor version (avoiding the parse of the package.json from the user).
This feature is only available on Yarn so when using NPM it'll be skipped.

2) Parsing the packages.json
This path works with both NPM and Yarn (used when user simply types `yarn install`), it'll look for any sibling reference that doesn't match the desired version or uses the `^` symbol on the sibling version.

This is a tricky code since it postinstall gets called before the package manager decides to update the packages.json .

Furthermore, this option will list all the required changes and also the required command in order to update the libraries (We could update the user's package.json but I think that's too evasive/extra effort by the script).

Ther's also a flag `--update-sentry-capacitor` to ignore the postinstall check since when updating a library using method 2, postinstall will only read the previous version prior the update, avoiding the user to be stuck on the update process).

Sample from what the user will see when installing the wrong sibling.

NPM
```
npm ERR!       throw IncompatibilityError;
npm ERR!       ^
npm ERR! This version of Sentry Capacitor is incompatible with the following installed packages:
npm ERR! @sentry/angular version ^7.13.0
npm ERR! @sentry/vue version ^7.13.0
npm ERR! Please install the mentioned packages with exactly with version 7.15.0 and with the argument --update-sentry-capacitor
npm ERR! npm install --save-exact @sentry/angular@7.15.0 @sentry/vue@7.15.0  --update-sentry-capacitor
```

Yarn
```
      throw IncompatibilityError;
      ^
This version of Sentry Capacitor is incompatible with the following installed packages:
@sentry/angular version 7.13.0
@sentry/vue version 7.13.0
Please install the mentioned packages with exactly with version 7.15.0 and with the argument --update-sentry-capacitor
yarn add --exact @sentry/angular@7.15.0 @sentry/vue@7.15.0  --update-sentry-capacitor
```

scripts/check-siblings.js
```diff
@@ -0,0 +1,145 @@
...
+      const capacitorPackageJson = fs.readFileSync(capacitorPackagePath, 'utf8');
+
+      const version = capacitorPackageJson.match(jsonFilter);
+      if (version && version.groups['version']) {
+            return version.groups['version'];
+      }
+      return undefined;
+}
+
+function validSentryPackageParameters(packages) {
```
kahest, line 52:
```suggestion
function ValidateSentryPackageParameters(packages) {
```

kahest: m: maybe also makes sense to add doc header to describe when exc is thrown

scripts/check-siblings.js
```diff
@@ -0,0 +1,145 @@
...
+      }
+}
+if (incompatiblePackages.length > 0) {
+      let IncompatibilityError = "This version of Sentry Capacitor is incompatible with the following installed packages: \n";
+      let packagesList = ''
+      for (const package of incompatiblePackages) {
+            IncompatibilityError += package[0] + ' version ' + package[1] + '\n';
+            packagesList += package[0] + '@' + siblingVersion + ' ';
+      }
+      IncompatibilityError += 'Please install the mentioned packages with exactly with version ' + siblingVersion + ' and with the argument ' + updateArgument + '\n';
```
kahest, line 142:
```suggestion
      IncompatibilityError += 'Please install the mentioned packages exactly with version ' + siblingVersion + ' and with the argument ' + updateArgument + '\n';
```

scripts/check-siblings.js
```diff
@@ -0,0 +1,145 @@
...
+
+      const version = capacitorPackageJson.match(jsonFilter);
+      if (version && version.groups['version']) {
+            return version.groups['version'];
+      }
+      return undefined;
+}
+
+function validSentryPackageParameters(packages) {
+      let errorMessages = "";
```
kahest, line 53: m: could collect messages in an array and join on throw

kahest (commented):
Thanks for this! 🚀 
LGTM, some minor suggestions from my side.

AbhiPrasad (commented):
Hey @lucas-zimerman sorry for the late review.

Want to see if we can maybe pivot on this strategy a little - can we try doing something like we do here? https://github.com/getsentry/sentry-electron/pull/584

This way we don't need a `postinstall`, we just rely on TypeScript errors.

lucas-zimerman:
> Hey @lucas-zimerman sorry for the late review.
> 
> Want to see if we can maybe pivot on this strategy a little - can we try doing something like we do here? [getsentry/sentry-electron#584](https://github.com/getsentry/sentry-electron/pull/584)
> 
> This way we don't need a `postinstall`, we just rely on TypeScript errors.

That would be great, but here, instead of throwing an error on the SDK path, we only duplicated errors from with @sentry/types

```
Error: node_modules/@sentry/angular/node_modules/@sentry/types/types/globals.d.ts:2:11 - error TS2451: Cannot redeclare block-scoped variable '__DEBUG_BUILD__'.

2     const __DEBUG_BUILD__: boolean;
            ~~~~~~~~~~~~~~~

  node_modules/@sentry/types/types/globals.d.ts:2:11
    2     const __DEBUG_BUILD__: boolean;
                ~~~~~~~~~~~~~~~
    '__DEBUG_BUILD__' was also declared here.


Error: node_modules/@sentry/types/types/globals.d.ts:2:11 - error TS2451: Cannot redeclare block-scoped variable '__DEBUG_BUILD__'.

2     const __DEBUG_BUILD__: boolean;
            ~~~~~~~~~~~~~~~

  node_modules/@sentry/angular/node_modules/@sentry/types/types/globals.d.ts:2:11
    2     const __DEBUG_BUILD__: boolean;
                ~~~~~~~~~~~~~~~
    '__DEBUG_BUILD__' was also declared here.



[ERROR] An error occurred while running subprocess ng.

        ng.cmd run app:build exited with exit code 1.
```

scripts/check-siblings.js
```diff
@@ -1,36 +1,145 @@
...
+const { env } = require('process');
+
+const updateArgument = '--update-sentry-capacitor';
+
+/**
+ * If user requested to ignore the post-install
+ * @return {Boolean} true if requested to skip the post-install check, false otherwise.
+ */
+function SkipPostInstall() {
+      if (env.npm_config_update_sentry_capacitor) {
```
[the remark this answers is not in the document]
lucas-zimerman: somehow the editor was ignoring this format, I fixed on a new commit.

kahest (approved):

scripts/check-siblings.js
```diff
@@ -0,0 +1,149 @@
...
+    return env.npm_package_dependencies__sentry_browser;
+  }
+
+  let capacitorPackagePath = '';
+  if (env.npm_package_json) {
+    // NPM.
+    capacitorPackagePath = env.npm_package_json;
+  }
+  else {
+    capacitorPackagePath = __dirname + '../Package.json';
```
piotr-cz, line 42: I think here should be `'../package.json'`  (lowercase) as here: https://github.com/getsentry/sentry-capacitor/blob/0.11.0/package.json

lucas-zimerman: Thanks for spotting this! it should be fixed on the version!

PR #301, opened 2023-01-24
Bump Javascript SDK 7.31.1

lucas-zimerman:
This PR Bumps the Javascript SDK Dependencies and also fixes the changes required in order to be compatible with it.


src/transports/native.ts
```diff
@@ -1,5 +1,6 @@
-import { BaseTransportOptions, Envelope, Transport } from '@sentry/types';
-import { makePromiseBuffer, PromiseBuffer } from '@sentry/utils';
+import type { BaseTransportOptions, Envelope , Transport } from '@sentry/types';
```
kahest, line 1:
```suggestion
import type { BaseTransportOptions, Envelope, Transport } from '@sentry/types';
```

PR #324, opened 2023-02-17
Ignore Sentry Electron from Sibling Checks

lucas-zimerman:
Sentry Electron uses a different versioning number so it should be ignored from the sibling check, otherwise users will have to ignore the sibling check order to use the SDK.


kahest (approved):
LGTM, thanks! 🚀

PR #363, opened 2023-05-01
Replace prepublishonly by prepack.

### LINKED ISSUE #360: Missing yarn prepack step
### Environment

I'm using yarn v3 and importing `@sentry/capacitor` as a Git dependency (this is because I am using my fork). 

### Steps to Reproduce

1. Forked this repo
2. In my package.json dependencies, I now have `"@sentry/capacitor": "borgoat/sentry-capacitor#head=yarn-2"`
3. `yarn install`

### Expected Result

I can still import and use `@sentry/capacitor`, but based on my fork.

### Actual Result

I am not getting a correctly packaged dependency. The `dist` folder is missing, and the library is therefore unusable.

### Proposed Fix

Configure  the `prepack` lifecycle step to perform the library build, instead of `prepublishOnly`. This is the best practice according to Yarn as far as I can tell [^1].

> **Note about `yarn pack`**: Although rarely called directly, `yarn pack` is a crucial part of Yarn. Each time Yarn has to fetch a dependency from a "raw" source (such as a Git repository), it will automatically run `yarn install` and `yarn pack` on it to know which are the files to use.

and also

> [...] the prepublish hook shouldn't have side effects. In particular don't transpile the package sources in prepublish [...]

If you agree, I can provide a quick PR to address this.

[^1]: https://yarnpkg.com/advanced/lifecycle-scripts

lucas-zimerman:
Fixes #360

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -84,3 +84,6 @@ yalc.lock
 android/gradle/wrapper/gradle-wrapper.properties
 example/ionic-angular/android/capacitor.settings.gradle
 example/ionic-angular/android/app/capacitor.build.gradle
+
+#yarn pack files
+sentry-capacitor-v*.tgz
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixes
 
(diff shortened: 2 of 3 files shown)


kahest (approved):
LGTM, thanks! 🚀

PR #365, opened 2023-05-02
add stale.yml

lucas-zimerman:
After reviewing some Issues I noticed that some old issues didn't have any stale warning.

So I just added the missing Github integration that Sentry uses on other Projects.

In this case, it's basically the same implementation as the one done on React Native: https://github.com/getsentry/sentry-react-native/blob/main/.github/workflows/stale.yml

#skip-changelog


kahest (approved):
LGTM, thanks! 🚀

PR #386, opened 2023-05-31
Add support to Capacitor 5

lucas-zimerman:
This PR improves the previous PR in regard to add support to Capacitor 5 by also not breaking users that are opting to stay on Capacitor 4.

This PR also includes a new Sample using Capacitor 5

Thanks again to @borgoat  for the initial Pull Request!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Features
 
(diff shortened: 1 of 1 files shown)


kahest (approved):
looking good, thanks!

PR #400, opened 2023-06-27
Removed promises and wizard from dependencies.

lucas-zimerman:
Following the same behaviour of other SDKs, I am moving the `Wizard` to the `devDependencies` and removing the package `promise` that has no impact on the final product.

Close PR #180


Zapominacz (approved):

kahest (approved):
LGTM, thanks 🚀

PR #423, opened 2023-07-30
fix: dependencies sentry

TartanLeGrand:
For fix the dependencies if we are in monorepo.

```
This version of Sentry Capacitor is incompatible with the following installed packages:
@sentry/tracing version ^7.60.1
@sentry/vue version ^7.60.1
Please install the mentioned packages exactly with version 7.54.0 and with the argument --update-sentry-capacitor
yarn add --exact @sentry/tracing@7.54.0 @sentry/vue@7.54.0  --update-sentry-capacitor
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,9 +38,9 @@
   "license": "MIT",
   "peerDependencies": {
     "@capacitor/core": ">=3.0.0",
-    "@sentry/angular": "7.56.0",
-    "@sentry/react": "7.56.0",
-    "@sentry/vue": "7.56.0"
+    "@sentry/angular": "^7.56.0",
+    "@sentry/react": "^7.56.0",
+    "@sentry/vue": "^7.56.0"
   },
   "peerDependenciesMeta": {
     "@sentry/angular": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -38,9 +38,9 @@
   "license": "MIT",
   "peerDependencies": {
     "@capacitor/core": ">=3.0.0",
-    "@sentry/angular": "7.56.0",
-    "@sentry/react": "7.56.0",
-    "@sentry/vue": "7.56.0"
+    "@sentry/angular": "^7.56.0",
+    "@sentry/react": "^7.56.0",
+    "@sentry/vue": "^7.56.0"
```
lucas-zimerman, line 43: I am aware that it may be annoying, but we planned to only use fixed versions, so instead, I would recommend changing those versions to "7.60.1" instead of "^7.56.0"

This may be mitigated in the future if we do more frequent releases bumping Sentry JavaScript

TartanLeGrand: But you don't follow the sentry official version when they are published

maxnowack: @lucas-zimerman what's the reason for this?

lucas-zimerman: The main reason is we want to make sure that the release of capacitor SDK works with the given Sentry Siblings releases.

Are you not able to install Sentry Capacitor version 0.12.2 with Sentry React 7.56.0?

TartanLeGrand: Okay, but in the semantic versioning the second digit is the minor version, if you don't need the features of the 7.62.x you can stay in 7.56.0 in your minimal dependencies and you still stay compatible with the nexts minor versions.

Jorgagu:
@lucas-zimerman any update ? :thinking:

TartanLeGrand:
How we can have 2 diff versions in a mono repo ?

maxnowack:
@krystofwoldrich thanks for taking care of this. This is a good explanation of why version pinning should be maintained.

What would make updating sentry packages easier on our end would be if all packages followed the same versioning scheme. For example, all packages in the [sentry-javascript](https://github.com/getsentry/sentry-javascript) monorepo have the same version number. sentry-capacitor is also used exclusively in the javascript context, but has a different versioning. This makes updating the packages a bit difficult at the moment, since you can't just bump the sentry-javascript version without first checking which is the corresponding sentry-capacitor version.
Maybe it would be an idea to include sentry-capacitor in the monorepo of sentry-javascript (if that makes sense).

PR #438, opened 2023-08-08
Fix error when serializing objects with circular referencing

mlostekk:
When trying to serialize an object like this

```typescript
const subObject: { sub1: string, sub2: any } = { sub1: 'subObjectValue', sub2: undefined }
const givenObject = { root1: 'some value', root2: subObject };
subObject.sub2 = givenObject;
```

one will get an error like this

```
    TypeError: Converting circular structure to JSON
        --> starting at object with constructor 'Object'
        |     property 'b' -> object with constructor 'Object'
        --- property 'd' closes the circle
        at JSON.stringify (<anonymous>)
```

This PR fixes this behavior. A "flattened" output can be seen in the added [unit test](https://github.com/mlostekk/sentry-capacitor/blob/0eb8660011376c315d80a33c6a2b1b77750e69f0/test/wrapper.test.ts#L632).

### DIFF
--- a/src/scope.ts
+++ b/src/scope.ts
@@ -1,6 +1,7 @@
 import { Scope } from '@sentry/core';
 import type { Breadcrumb, User } from '@sentry/types';
 
+import { convertToNormalizedObject } from './utils/normalize';
 import { NATIVE } from './wrapper';
 
 /**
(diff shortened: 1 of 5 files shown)


lucas-zimerman:
Hey thank you for opening the PR!
Sorry to ask for this change, but since recently this was addressed on React Native, could you also use the way it was solved on React Native?

Basically copy the normalise to util/normalise
https://github.com/getsentry/sentry-react-native/blob/9ee4637908664fd8c5c34d592589352f6112e102/src/js/utils/normalize.ts
and instead of calling `JSON.stringify(value);`

Additionally, we could also add the tests that were used here:
https://github.com/getsentry/sentry-react-native/blob/9ee4637908664fd8c5c34d592589352f6112e102/test/utils/normalize.test.ts

mlostekk:
@lucas-zimerman Not entirely sure if i covered all areas. But I have updated it accordingly to the RN solution

PR #444, opened 2023-08-17
Bump sentry javascript 7.64.0

lucas-zimerman:
Bumps Sentry JavaScript to 7.64.0

Minor changes:
- Removed a React Native flag that wasn't being used on the vscode Settings.
- Enforced Visual Studio Code to use the line ending mode LF and not CRLF by default

Lastly
it was disabled the esconfig `@sentry-internal/sdk/no-class-field-initializers` to allow integrations to initialize parameters outside of constructor (mostly the name of the integration)
The same change also applies to [React Native](https://github.com/getsentry/sentry-react-native/pull/3226))


kahest (approved):
LGTM, thanks! 🚀

Issue #424, opened 2023-07-31
iOS issue - OutOfMemory on app launch


beppemac:
### Environment

Angular 16
Capacitor 5
Ionic 7
@sentry/angular-ivy: 7.54.0
@sentry/capacitor: 0.12.1
@sentry/tracing: 7.54.0

### Steps to Reproduce

1. Following the installation guide for angular/capacitor set up
2. Run the iOS app

### Expected Result

The app launches without reporting errors

### Actual Result

The app launches without any problems for the user. 
But the error "OutOfMemory: The OS most likely terminated your app because it overused RAM." is logged on Sentry, without stack trace.

To see more informations about the error, I launched the app in debug mode, and below is the log from XCode console.
If I remove Sentry.init call, the error doesn't show up.

> =================================================================
Main Thread Checker: UI API called on a background thread: -[UIApplication delegate]
PID: 969, TID: 172129, Thread name: (none), Queue name: bridge, QoS: 0
Backtrace:
4   Sentry                              0x0000000105737c90 -[SentryUIViewControllerSwizzling swizzleRootViewControllerFromUIApplication:] + 52
5   Sentry                              0x0000000105736f3c -[SentryUIViewControllerSwizzling start] + 80
6   Sentry                              0x00000001056fe434 -[SentryPerformanceTrackingIntegration installWithOptions:] + 456
7   Sentry                              0x0000000105715bd4 +[SentrySDK installIntegrations] + 832
8   Sentry                              0x0000000105714208 +[SentrySDK startWithOptionsObject:] + 440
9   SentryCapacitor                     0x00000001050f4d18 $s15SentryCapacitorAAC13initNativeSdkyySo13CAPPluginCallCF + 608
10  SentryCapacitor                     0x00000001050f57b4 $s15SentryCapacitorAAC13initNativeSdkyySo13CAPPluginCallCFTo + 52
11  Capacitor                           0x0000000104dc7fd0 $s9Capacitor0A6BridgeC12handleJSCall4callyAA0D0V_tFyycfU0_ + 856
12  Capacitor                           0x0000000104dbc478 $sIeg_IeyB_TR + 48
13  libdispatch.dylib                   0x00000001054a85a8 _dispatch_call_block_and_release + 32
14  libdispatch.dylib                   0x00000001054aa05c _dispatch_client_callout + 20
15  libdispatch.dylib                   0x00000001054b210c _dispatch_lane_serial_drain + 988
[...]

Issue #207, opened 2022-08-08
UI API called on a background thread


arcdigital:
### Environment

Which SDK and version?
Capacitor 0.9.0
Vue 7.8.1
Tracing 7.8.1

### Steps to Reproduce

Create a Capacitor 4 + Vue 3 project and include
```
    "@sentry/capacitor": "0.9.0",
    "@sentry/integrations": "7.8.1",
    "@sentry/tracing": "7.8.1",
    "@sentry/vue": "7.8.1",
```

Enable tracing with 
`      tracesSampleRate: 1.0,`

### Actual Result

The console shows a ton of `Main Thread Checker: UI API called on a background thread` warnings. They don't appear if the SDK is initialized, but tracing isn't enabled.

```
Main Thread Checker: UI API called on a background thread: -[UIApplication delegate]
PID: 89405, TID: 6154204, Thread name: (none), Queue name: bridge, QoS: 0
Backtrace:
4   Sentry                              0x0000000101e9ce18 -[SentryUIViewControllerSwizzling swizzleRootViewControllerFromUIApplication:] + 56
5   Sentry                              0x0000000101e9c3bc -[SentryUIViewControllerSwizzling start] + 80
6   Sentry                              0x0000000101e7fd70 -[SentryPerformanceTrackingIntegration installWithOptions:] + 516
7   Sentry                              0x0000000101e88ed8 +[SentrySDK installIntegrations] + 892
8   Sentry                              0x0000000101e875dc +[SentrySDK startWithOptionsObject:] + 416
9   SentryCapacitor                     0x000000010198a86c $s15SentryCapacitorAAC13initNativeSdkyySo13CAPPluginCallCF + 636
10  SentryCapacitor                     0x000000010198b17c $s15SentryCapacitorAAC13initNativeSdkyySo13CAPPluginCallCFTo + 52
11  Capacitor                           0x00000001011e0b38 $s9Capacitor0A6BridgeC12handleJSCall4callyAA0D0V_tFyycfU_ + 816
12  Capacitor                           0x00000001011d639c $sIeg_IeyB_TR + 52
13  libdispatch.dylib                   0x0000000101fb033c _dispatch_call_block_and_release + 24
14  libdispatch.dylib                   0x0000000101fb1b94 _dispatch_client_callout + 16
15  libdispatch.dylib                   0x0000000101fb8cc4 _dispatch_lane_serial_drain + 764
[...]

PR #448, opened 2023-08-29
Fix: Always start Sentry Cocoa on Main Thread

lucas-zimerman:
We must guarantee that Sentry Cocoa initialises on the Main Thread, usually this is the expected flow that usually happens but there are cases where It'll be initialised on another thread and may lead to a Crash.
This fix guarantees that Sentry Cocoa is always initialised on the main thread

Fixes: #424, #443, #207

Lastly, added the missing pod file for one of the samples, it has no impact on the main code and it's always generated when building.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- (iOS) UI API called on a background thread ([#448](https://github.com/getsentry/sentry-capacitor/pull/448))
+
 ## 0.12.3
 
 ### Fixes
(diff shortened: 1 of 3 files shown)


brustolin:
> @brustolin do you have a concern with the Cocoa SDK being initialized in this way?

Its possible for the SDK to loose automatic tracing of the first view controller being displayed. But I dont know if capacitor uses cocoa automatic view controller performance tracking.

lucas-zimerman:
> > @brustolin do you have a concern with the Cocoa SDK being initialized in this way?
> 
> Its possible for the SDK to loose automatic tracing of the first view controller being displayed. But I dont know if capacitor uses cocoa automatic view controller performance tracking.
> 
> Otherwise, there is no problem at all.

From what I saw with and without this change there wasn't any differences there weren't any changes to the sent transaction/events.

PR #454, opened 2023-09-01
Fix: enableWatchdogTerminationTracking (replaces enableOutOfMemoryTracking) is now properly set on iOS.

### LINKED ISSUE #445: TypeScript/hybrid layer not implementing new enableWatchdogTerminationTracking setting
### Environment

Upgrading from version 0.11.4 to version 0.12.2 of Sentry Capacitor

### Steps to Reproduce

1. Prior to upgrading to v0.12.2, our app implemented a [setting](https://docs.sentry.io/platforms/javascript/guides/capacitor/configuration/options/#enableOutOfMemoryTracking) called `enableOutOfMemoryTracking`
2. This setting is used to disable events sent to Sentry by the native layer if the OS kills the native app on iOS.
3. We are setting it to `false` to disable [out of memory](https://docs.sentry.io/platforms/apple/guides/ios/configuration/out-of-memory/) tracking on iOS.

### Expected Result

We expected setting `enableOutOfMemoryTracking: false` to disable out of memory tracking on iOS.

### Actual Result

Out of memory tracking still occurs on iOS.

It seems this setting was [renamed](https://docs.sentry.io/platforms/apple/guides/ios/configuration/watchdog-terminations/) to `enableWatchdogTerminationTracking` with version 8.0.0 of Sentry Apple SDK. This might be why configuring the setting `enableOutOfMemoryTracking: false` no longer seems to disable it.

lucas-zimerman:
New implementation of Sentry Cocoa uses `enableWatchdogTerminationTracking` instead of  `enableOutOfMemoryTracking `
This PR fixes the correct value to be sent to the Native SDK, allowing both methods to be used, and also deprecates enableOutOfMemoryTracking to warn users that this feature will be removed on a major version.

Fixes #445

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- enableWatchdogTerminationTracking (replaces enableOutOfMemoryTracking) is now properly set on iOS. ([#454](https://github.com/getsentry/sentry-capacitor/pull/454))
+
 ## 0.12.3
 
 ### Fixes
--- a/src/nativeOptions.ts
+++ b/src/nativeOptions.ts
@@ -16,7 +16,8 @@ export function FilterNativeOptions(options: CapacitorOptions): CapacitorOptions
     dsn: options.dsn,
     enabled: options.enabled,
     enableNdkScopeSync: options.enableNdkScopeSync,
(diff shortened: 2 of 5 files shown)


kahest (approved):
LGTM, thanks!
@brustolin please take a quick look as a sanity check for the usage on the Cocoa SDK

brustolin (approved):
LGTM

Issue #405, opened 2023-07-05
Peer Dependencies npm install --save @Sentry/capacitor Fails


15aabruzzese:
### Environment

Angular 16: I can successfully download @sentry/angular-ivy 
Capacitor 5:
Iconic 7:
npm 9.5.1
node 18.16

### Steps to Reproduce

1. Following the installation guide for angular/capacitor set up
2. `npm install --save @sentry/capacitor @sentry/angular-ivy` or `npm install --save @sentry/capacitor` or `npm install --save @sentry/capacitor --force`
3.  installation fails

![image](https://github.com/getsentry/sentry-capacitor/assets/96092264/32512369-1b63-4750-860e-e0ae0e40ca46)

### Expected Result
The npm install would be successful

### Actual Result
install fails (see screenshots)

Current Dependencies: 
"dependencies": {
    "@angular/common": "^16.0.3",
    "@angular/core": "^16.0.3",
    "@angular/forms": "^16.0.3",
    "@angular/platform-browser": "^16.0.3",
    "@angular/platform-browser-dynamic": "^16.0.3",
    "@angular/router": "^16.0.3",
    "@awesome-cordova-plugins/app-version": "^6.3.0",
    "@capacitor-community/firebase-analytics": "^5.0.1",
    "@capacitor/android": "^5.0.4",
    "@capacitor/app": "^5.0.2",
    "@capacitor/camera": "^5.0.2",
    "@capacitor/core": "^5.0.5",
    "@capacitor/haptics": "^5.0.2",
    "@capacitor/ios": "^5.0.4",
    "@capacitor/keyboard": "^5.0.2",
    "@capacitor/status-bar": "^5.0.2",
    "@ionic/angular": "^7.0.9",
    "@ionic/pwa-elements": "^3.1.1",
    "@ionic/storage-angular": "^4.0.0",
    "@trapezedev/configure": "^7.0.10",
    "capacitor-native-settings": "^5.0.0",
    "cordova-plugin-app-version": "^0.1.14",
    "envsub": "^4.1.0",
    "ionicons": "^7.1.1",
    "ngx-mask": "^16.0.8",
    "rxjs": "^7.8.1",
    "swiper": "^8.4.6",
    "tslib": "^2.5.2",
    "zone.js": "^0.13.0"
  },
  "devDependencies": {
    "@angular-devkit/build-angular": "^16.0.3",
    "@angular-eslint/builder": "^16.0.3",
[...]

PR #458, opened 2023-09-05
Fix for Capacitor 5/Angular 16 peer dependencies error

ahunter135:


afilp (approved):

lucas-zimerman:
In this case wouldn't it break for uses using the non ivy version of Angular?

ahunter135:
@lucas-zimerman 

It might. I'm new to this. But we also need to do an install and it fails without this. 

Side effect of this change: I'm on capacitor angular, but I can't import it from there. I have to import from sentry/angular-ivy. So there could be more to do here. 

I just wanted to start the discussion back up because at this point, sentry is not supporting capacitor 5 and angular 16.

emelampianakis:
angular 16 needs the ivy distribution. if this is the case maby you can remove the dependency completely so each dev can install the package needed

ahunter135:
So instead of putting ivy in, just remove it entirely? I can do that. @emelampianakis

emelampianakis:
> So instead of putting ivy in, just remove it entirely? I can do that. @emelampianakis

Yeah, I think this is the correct approach, but then the Installation part of the readme should also be updated

emelampianakis:
Hello @ahunter135 can you try the changes I mentioned above?
Just remove both @sentry/angular & @sentry/angular-ivy dependencies and update the readme file.
We really need this fix cause as it is right now angular 16 projects cant use this package.

ahunter135:
@emelampianakis made the requested changes!

miguel-S73 (approved):

afilp:
@lucas-zimerman Can you please approve and merge the changes so that we can start using the fix soon (with Angular 16)? Thanks a lot!

krystofwoldrich:
Based on https://github.com/getsentry/sentry-capacitor/issues/405#issuecomment-1732593860 the peer deps are not the issue.

Thank you, everyone, for the comments, please use `@sentry/angular-ivy` with Angular 12 and newer.

Follow up PR with readme update from this PR -> https://github.com/getsentry/sentry-capacitor/pull/469

PR #469, opened 2023-09-25
Add sentry-angular-ivy to optinal deps

krystofwoldrich:
- related to https://github.com/getsentry/sentry-capacitor/issues/405
- supersedes https://github.com/getsentry/sentry-capacitor/pull/458

Capacitor SDK uses peer deps to enforece supported version of the related Sentry Javascript packages. This is important as mismatched versions can lead to 
unexpected behaviour, mismatched types etc.

Currently the latest angular-ivy package was not included. This pr adds it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,11 +40,15 @@
   "license": "MIT",
   "peerDependencies": {
     "@capacitor/core": ">=3.0.0",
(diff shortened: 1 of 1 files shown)


lucas-zimerman (approved):
Thanks for the PR!

PR #492, opened 2023-10-31
chore: Use sentry-internal/typescript instead of sentry/typescript

jcesarmobile:
A fresh npm install has peer dependency issues because of a conflict in the typescript version and the `@sentry/typescript` version.
I saw that it was renamed long ago to `sentry-internal/typescript`, so I've updated the plugin to use that instead and also fix the issues generated from the package change.


lucas-zimerman:
Sorry for this issue and thanks for the PR!

lucas-zimerman (approved):
Thank you for the PR!

Issue #440, opened 2023-08-10
Hostnames other than localhost break stack traces


jskrepnek:
The RewriteFrames integration assumes that the hostname is `localhost`, but it is normal to override this for various reasons.

https://github.com/getsentry/sentry-capacitor/blob/32dc319589b6bc6792de95e25eb073047b3f14d1/src/sdk.ts#L51-L53

jskrepnek:
👋

Issue #391, opened 2023-06-14
Support frame rewriting for migrated Cordova apps using ionic scheme


jskrepnek:
Apps that migrated from Cordova to Capacitor may still be using `ionic` as the server scheme.  The default integration that [rewrites frames](https://github.com/getsentry/sentry-capacitor/blob/main/src/sdk.ts#L50) does not take this into account, resulting in invalid frame urls .  The only recourse is to apply custom processing before events are sent to fix the filenames.

I think it makes sense to bake support for the `ionic` scheme into the plugin, but an alternative could be to provide a configuration point for the removal of custom schemes and protocols.

lucas-zimerman:
By any chance do you have any sample/example of the affected stacktrace?

jskrepnek:
Hi @lucas-zimerman 

Here's an example:

```
at testWebUnhandledError(app:///ionic://localhost/7214.a1b610e6cad107f9.js:1:42273)
  at Df(app:///ionic://localhost/main.569f5323c6e54cbf.js:1:3193112)
  at be(app:///ionic://localhost/main.569f5323c6e54cbf.js:1:3193288)
  at ? (app:///ionic://localhost/main.569f5323c6e54cbf.js:1:3297173)
  at rt(app:///ionic://localhost/main.569f5323c6e54cbf.js:1:2440127)
  at onInvokeTask(app:///ionic://localhost/main.569f5323c6e54cbf.js:1:3230564)
  at ? (app:///ionic://localhost/polyfills.ced8424825c55a82.js:1:3466)
  at ? (app:///ionic://localhost/polyfills.ced8424825c55a82.js:1:9604)
  at l(app:///ionic://localhost/polyfills.ced8424825c55a82.js:1:23295)
  at c(app:///ionic://localhost/polyfills.ced8424825c55a82.js:1:23678)
```

lucas-zimerman:
Ok, thank you for the sample, I'll add this to the queue of features to be done!

jskrepnek:
Seems like the bot is not working

jskrepnek:
don't close

PR #494, opened 2023-11-01
fix: correct the capacitor url replacement

jcesarmobile:
Instead of hardcoding the replacement of the Capacitor url for just `https://localhost` and `capacitor://localhost`, use the value on `window.WEBVIEW_SERVER_URL` that Capacitor sets with the proper value based on the user scheme and hostname configuration.

I had to disable some eslint rules for the line, let me know if there is a better approach

closes https://github.com/getsentry/sentry-capacitor/issues/440
closes https://github.com/getsentry/sentry-capacitor/issues/391

### DIFF
--- a/src/sdk.ts
+++ b/src/sdk.ts
@@ -11,6 +11,7 @@ import type { CapacitorOptions } from './options';
 import { CapacitorScope } from './scope';
 import { DEFAULT_BUFFER_SIZE, makeNativeTransport } from './transports/native';
 import { makeUtf8TextEncoder } from './transports/TextEncoder';
+import { getCurrentServerUrl } from './utils/webViewUrl';
 import { NATIVE } from './wrapper';
 
 /**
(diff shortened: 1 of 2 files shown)


src/sdk.ts
```diff
@@ -46,12 +46,10 @@ export function init<O>(
...
           const isReachableHost = /^https?:\/\//.test(frame.filename);
-
           frame.filename = frame.filename
-            .replace(/^https?:\/\/localhost(:\d+)?/, '')
-            .replace(/^ng:\/\//, '')
-            .replace(/^capacitor:\/\/localhost(:\d+)?/, '');
-
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-explicit-any, no-restricted-globals
+            .replace((window as any).WEBVIEW_SERVER_URL, '')
+            .replace(/^ng:\/\//, '');
```
lucas-zimerman, line 52: would the following work for you?

```typescript
interface GlobalObject  extends Window{
  WEBVIEW_SERVER_URL?: string;
};

/**
 * Return the current webview url if found.
 */
export function getCurrentServerUrl(): string | undefined {
    return (self as GlobalObject).WEBVIEW_SERVER_URL;
}

```
this code could be added on src/tools/webViewUrl.ts

and just as a safe rule, we could use the old rule
```typescript
            .replace(/^https?:\/\/localhost(:\d+)?/, '')
            .replace(/^capacitor:\/\/localhost(:\d+)?/, '');
            .replace(/^ng:\/\//, '')
```
if the return of getCurrentServerUrl was undefined, else we could use the new method with `.replace(/^ng:\/\//, '')`.

lucas-zimerman:
THank you for the PR!

src/sdk.ts
jcesarmobile: That could work too. But my sentry trial expired, so I can’t test the proposed changes.

lucas-zimerman: I have tested it and it is working, if you prefer I can merge this PR to another branch and do the final changes before going to the main branch

jcesarmobile: despite the trial ended, I could still test with the free plan, I've updated with the requested changes

lucas-zimerman (approved):

PR #499, opened 2023-11-10
Bump: Sentry/Javascript  7 80 0 + minor fixes

### LINKED ISSUE #497: `sentry-capacitor` pinned to `sentry-javascript` version with sourcemap bug
### Environment

0.14.0

### Steps to Reproduce
Same as issue filed in [9378](https://github.com/getsentry/sentry-javascript/issues/9378).

### Expected Result
I expected to be able to bump the version of my `sentry-javascript` to use the latest patch.

### Actual Result
`sentry-capacitor` is [explicitly pinned](https://github.com/getsentry/sentry-capacitor/blob/main/package.json#L63-L69) to the version with the bug.

lucas-zimerman:
This PR bumps Sentry/Javascript to the latest version 7.80.0 and also includes some minor fixes.

Minor Fixes:
- Add changelog for #494
- Added tests for #494
- Moved RewriteFames outside of sdk.ts to rewriteframs.ts on integrations.
- Fixed Wizard  missing sentry/cli on samples (https://github.com/getsentry/sentry-wizard/issues/483)

Fixes #497


.gitignore
```diff
@@ -87,3 +87,6 @@ example/ionic-angular/android/app/capacitor.build.gradle
 
 #yarn pack files
 sentry-capacitor-v*.tgz
+
+#sentry tokens
+*.sentryclirc
```
krystofwoldrich, line 92: L: Do we have some config files with prefixes?

krystofwoldrich (approved):
Looks good. It's good that we keep the JS quite recent. 🚀

.gitignore
lucas-zimerman: if you run the sourcemap scripts from wizard it automatically adds the token files on the sample projects, so better be safe than leaking personal tokens

PR #512, opened 2023-11-24
Ref: Github Action for updating Sentry dependencies.

### LINKED ISSUE #510: Add update deps to capacitor
I noticed we're not getting auto update for dependencies.

We need somethjing like this here: https://github.com/getsentry/sentry-react-native/blob/main/.github/workflows/update-deps.yml

lucas-zimerman:
Close #510

#skip-changelog

### DIFF
--- a/.github/workflows/update-deps.yml
+++ b/.github/workflows/update-deps.yml
@@ -1,0 +1,57 @@
+name: Update Dependencies
+
+on:
+  # Run every day.
+  schedule:
+    - cron: '0 3 * * *'
+  # And on on every PR merge so we get the updated dependencies ASAP, and to make sure the changelog doesn't conflict.
+  push:
+    branches:
+      - main
+
+jobs:
+  android:
+    uses: getsentry/github-workflows/.github/workflows/updater.yml@v2
+    with:
+      path: scripts/update-android.sh
+      name: Android SDK
+      pr-strategy: update
+    secrets:
+      api-token: ${{ secrets.CI_DEPLOY_KEY }}
+
+  cocoa:
+    uses: getsentry/github-workflows/.github/workflows/updater.yml@v2
+    with:
+      path: scripts/update-cocoa.sh
+      name: Cocoa SDK
+      pr-strategy: update
+    secrets:
+      api-token: ${{ secrets.CI_DEPLOY_KEY }}
+
+  javascript:
+    uses: getsentry/github-workflows/.github/workflows/updater.yml@v2
+    with:
+      path: scripts/update-javascript.sh
+      name: JavaScript SDK
+      pr-strategy: update
+    secrets:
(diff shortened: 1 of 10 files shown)


scripts/update-javascript.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+
+tagPrefix=''
+repo="https://github.com/getsentry/sentry-javascript.git"
+packages=('@sentry/browser' '@sentry/core' '@sentry/integrations' '@sentry/react' '@sentry/vue' '@sentry/angular' '@sentry/angular-ivy' '@sentry/types' '@sentry/utils')
```
krystofwoldrich, line 7: I believe this won't work for `@sentry/angular-ivy`, `@sentry/angular`, `@sentry/react`, and `@sentry/vue`. As the `update-package-json.sh` runs `yarn upgrade`. 

I tried to run `yarn upgrade` and it added these as deps not upgrading the peers.

scripts/update-cli.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+tagPrefix=''
+repo="https://github.com/getsentry/sentry-cli.git"
+packages=('@sentry/cli')
+
+. $(dirname "$0")/update-package-json.sh
```
krystofwoldrich, line 8: Sentry-Capacitor doesn't have Sentry CLI as a dependency. Do we want to change that or otherwise we should remove this update script?

.github/workflows/update-deps.yml
```diff
@@ -0,0 +1,57 @@
+name: Update Dependencies
+
+on:
+  # Run every day.
+  schedule:
+    - cron: '0 3 * * *'
+  # And on on every PR merge so we get the updated dependencies ASAP, and to make sure the changelog doesn't conflict.
+  push:
+    branches:
+      - main
```
krystofwoldrich, line 10: We can TMP add this branch here to check if the scripts work as expected.

scripts/update-javascript.sh
lucas-zimerman: nice catch!, sucks that yarn upgrade doesn't support peerDependencies...
I added another way of updating the peerDependencies by using the add command with the --peer flag instead of relying on the upgrade command

lucas-zimerman:
The PR will not pass for the time being because the update logic compares with the main branch and this branch isn't the main branch, but all the previous steps are passing

.github/workflows/update-deps.yml
```diff
@@ -0,0 +1,59 @@
...
+      # Temporary for testing.
+      - feat/auto-bump=dep
```
lucas-zimerman, line 12:
```suggestion
```

kahest (approved):
LGTM, thanks for adding this! 🚀

PR #551, opened 2024-01-24
fix samples with correct sentry version, add client

lucas-zimerman:


lucas-zimerman:
PR Replaced by https://github.com/getsentry/sentry-capacitor/pull/573

PR #552, opened 2024-01-24
Feat: Add support of client report

lucas-zimerman:
This PR implements the support of Client Report to Sentry Capacitor

#skip-changelog


src/client.ts
```diff
@@ -10,12 +12,14 @@ import type {
...
   UserFeedback,
 } from '@sentry/types';
-import { logger, SentryError } from '@sentry/utils';
+import { dateTimestampInSeconds, logger, SentryError } from '@sentry/utils';
 
 import type { CapacitorClientOptions } from './options';
 import { mergeOutcomes } from './utils/outcome';
 import { NATIVE } from './wrapper';
 
+export const EnvelopeitemsIndex = 1;
```
krystofwoldrich, line 21: L: I would not export this if used only in the client file.

M: Missing capital I.
```suggestion
export const EnvelopeItemsIndex = 1;
```

src/client.ts
```diff
@@ -163,6 +166,20 @@ export class CapacitorClient extends BaseClient<CapacitorClientOptions> {
...
+    if (outcomes.length > 0) {
+      const clientReportItem: ClientReportItem = [
+        { type: 'client_report' },
+        {
+          timestamp: dateTimestampInSeconds(),
+          discarded_events: outcomes,
+        },
+      ];
+
+      envelope[EnvelopeitemsIndex].push(clientReportItem);
```
krystofwoldrich, line 182:
```suggestion
      envelope[EnvelopeItemsIndex].push(clientReportItem);
```

test/client.test.ts
```diff
@@ -294,35 +303,35 @@ describe('Tests CapacitorClient', () => {
...
-      const actualEvent: Event | undefined = <Event>(
-        mockedSend.mock.calls[0][firstArg][envelopeItems][0][envelopeItemPayload]
-      );
-      expect(actualEvent?.sdk?.packages).toEqual([
-        {
-          name: SDK_PACKAGE_NAME,
-          version: SDK_VERSION,
-        },
-      ]);
+  /* TODO: Fix SDKInfo
```
krystofwoldrich, line 306: L: From what I see this is unrelated to the PR, looks like only the indentation changed. 

What needs to be fixed with SDKInfo?

krystofwoldrich (approved):
Looks good, the same impl as RN, 
I have noticed one small typo, otherwise, it's good to go! 🚀

test/client.test.ts
lucas-zimerman: Yeah only the indentation changed.

I made a PR only for SDKInfo separately just to avoid fixing everything on a single PR.

It's basically missing the default export from the SDK on the integration.

PR #572, opened 2024-02-12
Fix: RewriteFrames on SDK Bump

lucas-zimerman:
This PR does the following tasks:

- Fixed break change with Rewrite frames.
- Refactored and relocated most of the tests from RewriteFrames to an unique file, where the sdk test will only validate the integration is being added.
- Added a new "broken" node library sample for testing stack traces coming from a node module.

#skip-changelog (it will be included on the main PR).
Sub-Task of #573


example/broken_node_module/src/hello-bug.ts
```diff
@@ -0,0 +1,8 @@
+export function throw_error() {
+  // Throw a new error with a custom message
+  throw new Error('This is an not intentional error from the library');
```
krystofwoldrich, line 3: Is this intentional error or not? The message says not but it looks like it is?

package.json
```diff
@@ -23,11 +23,10 @@
     "swiftlint": "node-swiftlint",
     "prepack": "yarn run build",
     "postinstall": "node scripts/check-siblings.js",
-    "bump:v5": "cd example/ionic-angular-v5 && yalc add @sentry/capacitor && yarn install && ionic build && npx cap sync",
-    "bump:v3": "cd example/ionic-angular && yalc add @sentry/capacitor && yarn install && ionic build && npx cap sync",
-    "bump:v2": "cd example/ionic-angular-v2 && yalc add @sentry/capacitor && yarn install && ionic build && npx cap sync",
-    "bump:sample-vue": "cd example/ionic-vue3 && yalc add @sentry/capacitor && yarn install && ionic build && npx cap sync",
```
krystofwoldrich, line 29: Why is the `sample-vue` removed?

lucas-zimerman: Nice catch, something went wrong on the merge from the main branch

src/integrations/rewriteframes.ts
```diff
@@ -12,17 +12,25 @@ export function createCapacitorRewriteFrames(): Integration {
   const rewriteFrames = new RewriteFrames({
     iteratee: (frame: StackFrame) => {
       if (frame.filename) {
+
+        if (frame.platform === 'java' || frame.platform === 'cocoa') {
+          // Because platform is not required in StackFrame type
+          // we assume that if not set it's javascript
+          return frame;
+        }
```
krystofwoldrich, line 20: Does this apply in Capacitor?

In React Native this comes from https://github.com/getsentry/sentry-react-native/blob/66cd5b29cd61243789bad98b33f1364df1681510/src/js/integrations/nativelinkederrors.ts#L31

lucas-zimerman: Capacitor doesn't have Native Linked errors so I'll remove this code.

test/integrations/rewriteframes.test.ts
```diff
@@ -0,0 +1,579 @@
...
+      },
+    };
+
+    const event = processEvent(SENTRY_JVM_EXCEPTION_EVENT) as Event;
+    expect(event.exception?.values?.[0]).toEqual(EXPECTED_SENTRY_JVM_EXCEPTION);
+  });
+
+
+  it('should parse Capacitor errors on Android', async () => {
+    const ANDROID_capacitor_NATIVE = {
```
krystofwoldrich, line 85:
```suggestion
    const ANDROID_CAPACITOR_NATIVE = {
```

test/integrations/rewriteframes.test.ts
```diff
@@ -0,0 +1,579 @@
...
+        'i@index.android.bundle:176:358\n' +
+        'i@index.android.bundle:93:90\n' +
+        'u@index.android.bundle:93:150\n' +
+        '_receiveRootNodeIDEvent@index.android.bundle:156:544\n' +
+        'receiveTouches@index.android.bundle:156:918\n' +
+        'value@index.android.bundle:29:3016\n' +
+        'index.android.bundle:29:955\n' +
+        'value@index.android.bundle:29:2417\n' +
+        'value@index.android.bundle:29:927\n' +
+        '[native code]',
```
krystofwoldrich, line 212: This looks like RN stack trace, should we update it and add the actual capacitor stack?

test/integrations/rewriteframes.test.ts
```diff
@@ -0,0 +1,579 @@
...
+    ]])
+    ('should parse Angular, Capacitor and Native errors %s', async (_, frame, expectedFrame) => {
+
+      const error = { values: [{ stacktrace: { frames: [frame] } }] };
+      const event = await processEvent({ exception: error }) as Event;
+      const firstFrame = (firstStackTraceFromEvent(event) as Stacktrace).frames?.pop();
+      expect(firstFrame).toMatchObject(expectedFrame);;
+    });
+
+  it('should do something with the mocked url', async () => {
```
krystofwoldrich, line 534:
```suggestion
  it('should remove current server url from frame filename', async () => {
```

test/integrations/rewriteframes.test.ts
```diff
@@ -0,0 +1,579 @@
+jest.mock('../../src/utils/webViewUrl');
+
+import type { Exception } from '@sentry/browser';
+import { defaultStackParser, eventFromException } from '@sentry/browser';
+import type { Client, Event, EventHint, StackFrame, Stacktrace } from '@sentry/types';
+
+import { createCapacitorRewriteFrames } from '../../src/integrations/rewriteframes';
+import * as webViewUrl from '../../src/utils/webViewUrl';
+
+describe('RewriteFrames', () => {
```
krystofwoldrich, line 10: Can we add a dev and release iOS examples?

src/integrations/rewriteframes.ts
```diff
@@ -12,17 +12,25 @@ export function createCapacitorRewriteFrames(): Integration {
...
-            .replace(/^capacitor:\/\/localhost(:\d+)?/, '');
-        }
+          frame.filename = frame.filename
+            .replace(/^https?:\/\/localhost(:\d+)?/, '')
+            .replace(/^capacitor:\/\/localhost(:\d+)?/, '')
+          }
         frame.filename = frame.filename.replace(/^ng:\/\//, '');
 
-        const isNativeFrame = frame.filename === '[native code]' || frame.filename === 'native';
+        const isNativeFrame = frame.filename === '[native code]' || frame.filename === 'native' || frame.filename.endsWith('.kt') ;
```
krystofwoldrich, line 26: When does `|| frame.filename.endsWith('.kt')` happen? 
We should add a test for it.

Is there something similar on iOS? Like `.mm`?

lucas-zimerman: To avoid implementing new things on a bump PR I'll rollback this code and if needed in the future we'll implement it back.

test/integrations/rewriteframes.test.ts
lucas-zimerman: the stacktrace for ios is the same from the tests so I'll skip it, but I added a test for dev builds

src/integrations/rewriteframes.ts
```diff
@@ -17,9 +17,10 @@ export function createCapacitorRewriteFrames(): Integration {
         if (serverUrl) {
           frame.filename = frame.filename.replace(serverUrl, '');
         } else {
-          frame.filename = frame.filename.replace(/^https?:\/\/localhost(:\d+)?/, '')
-            .replace(/^capacitor:\/\/localhost(:\d+)?/, '');
-        }
+          frame.filename = frame.filename
+            .replace(/^https?:\/\/localhost(:\d+)?/, '')
+            .replace(/^capacitor:\/\/localhost(:\d+)?/, '')
+          }
```
krystofwoldrich, line 23:
```suggestion
        }
```

PR #573, opened 2024-02-15
Bump: JavaScript 7.93.0

lucas-zimerman:
This PR Bumps Sentry JavaScript to version 7.93
There are internal break changes that required some code change but it shouldn't have any impact on the user side.
Before approval, this PR depends on #572 to fix the rewriteframes tests.


CHANGELOG.md
```diff
@@ -11,6 +19,9 @@
 - Bump Cocoa SDK from v8.8.0 to v8.18.0 ([#526](https://github.com/getsentry/sentry-capacitor/pull/526))
   - [changelog](https://github.com/getsentry/sentry-cocoa/blob/main/CHANGELOG.md#8180)
   - [diff](https://github.com/getsentry/sentry-cocoa/compare/8.8.0...8.18.0)
+- Bump JavaScript SDK from v7.81.1 to v7.93.0 ([#529](https://github.com/getsentry/sentry-capacitor/pull/529))
+  - [changelog](https://github.com/getsentry/sentry-javascript/blob/develop/CHANGELOG.md#7930)
+  - [diff](https://github.com/getsentry/sentry-javascript/compare/7.81.1...7.93.0)
```
krystofwoldrich, line 24:
```suggestion
```

scripts/update-javascript.sh
krystofwoldrich, line 1: GH says empty file, should we remove it, is it used somewhere?

src/integrations/eventorigin.ts
```diff
@@ -15,8 +15,8 @@ export class EventOrigin implements Integration {
   /**
    * @inheritDoc
    */
-  public setupOnce(addGlobalEventProcessor: (e: EventProcessor) => void): void {
-    addGlobalEventProcessor(event => {
```
krystofwoldrich, line 19: Just note, the callback is still `addGlobvalEventProcessor`. 

https://github.com/getsentry/sentry-javascript/blob/a7097d9ba2a74b2cb323da0ef22988a383782ffb/packages/core/src/integration.ts#L134

test/testutils.ts
krystofwoldrich, line 1: This file is not used anywhere.

test/tracing/utils.ts
krystofwoldrich, line 1: This file is not used.

scripts/update-javascript.sh
lucas-zimerman: there were 0 lines changed, not that it's empty.
I'll see if I can clear it on the next commit

scripts/update-javascript.sh
```diff
@@ -9,3 +9,4 @@ packages=('@sentry/browser' '@sentry/core' '@sentry/integrations' '@sentry/types
 packages+=('@sentry/hub' '@sentry/tracing')
 packages+=('@sentry-internal/eslint-config-sdk' '@sentry-internal/eslint-plugin-sdk' '@sentry-internal/typescript')
 . $(dirname "$0")/update-package-json.sh
+
```
lucas-zimerman, line 12:
```suggestion
```

src/integrations/eventorigin.ts
lucas-zimerman: Would you rather rollback this change and add the deprecated mark?

CHANGELOG.md
krystofwoldrich: This entry changes the 0.15.1 changelog, we should remove it, and only keep the changes in the unreleased.

scripts/update-javascript.sh
krystofwoldrich: I see the file permissions changed, was that intentional?

src/integrations/eventorigin.ts
krystofwoldrich: I would rather rollback in this case because the global in the name is more accurate, I believe the lint won't complain here, as the callback in not imported like in the other integrations.

krystofwoldrich (approved):
LGTM 🚀 

I have only one nit about the `eventorigin` integration callback name, I'll leave that up to you.

PR #584, opened 2024-02-28
fix: disable native sdk if DSN is not provided

Floppy012:
When no DSN is provided, the `initNativeSdk` function returns before the `enableNative` option is set. Effectively rendering the option useless.

This causes, the native sdk to still perform the checks of whether the native client is available and if that is not the case, then throws the `NativeClientError`.

A simple reproduction example would be:

```typescript
Sentry.init({
  dsn: undefined,
  enableNative: false,
});

setUser(null);
```

the `setUser` call will throw the NativeClientError.

I chose to set `enableNative` to false regardless of the provided option since the warning message that gets logged when the DSN is not provided states "... Native SDK will also not be initialized.", which should be enough of a hint that the native sdk will not function regardless of the provided `enableNative` unless a DSN is provided.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Native integration would not properly disable when no DSN was provided ([#584](https://github.com/getsentry/sentry-capacitor/pull/584))
+
 ## 0.16.0
 
 ### Features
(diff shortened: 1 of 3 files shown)


lucas-zimerman (approved):
Thank you for the PR!
We'll just finish releasing a new version and then we'll merge this PR

Floppy012:
Rebased & Fixed conflicts

PR #585, opened 2024-02-29
Feat: Add enable capture failed requests

lucas-zimerman:
This flag is enabled by default on iOS where it should be defaulted by false.

This flag will reduce error noises from clients since it'll stop recording Http errors that came from the Native side.
Users should only enable this flag if they do made Native code that does Http Requests.

TEST:

When this flag wasn't set, if a native Http Request error happened, the native SDK creates an event for it.
https://sentry-sdks.sentry.io/issues/5025177828/events/959b6258d9eb4931b196139436996011/?project=5522756

When disabled (the new behaviour), the same event will not be created.

### DIFF
--- a/src/options.ts
+++ b/src/options.ts
@@ -61,5 +61,4 @@ export interface CapacitorOptions
    * @default false
    */
   enableCaptureFailedRequests?: boolean;
-
 }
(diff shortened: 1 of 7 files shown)


src/options.ts
```diff
@@ -54,4 +54,12 @@ export interface CapacitorOptions
...
   * */
   enableOutOfMemoryTracking?: boolean;
+
+  /**
+   * When enabled, Sentry will capture failed XHR/Fetch requests. This option also enabled HTTP Errors on iOS.
+   *
+   * @default false
+   */
+  enableCaptureFailedRequests?: boolean;
+
```
lucas-zimerman, line 64:
```suggestion
```

krystofwoldrich (approved):
LGTM 🚀 Thank you!

PR #602, opened 2024-03-18
Feat: Add enableAppHangTracking and appHangTimeoutInterval.

### LINKED ISSUE #596: Disabling App Hang tracking does not work
### Environment

- Framework: Capacitor v4
- Platform: iOS

Dependencies

- @Sentry/capacitor: 0.16.0
- Sentry/HybridSDK: 8.18.0

### Steps to Reproduce

1. Initialize Sentry SDK in `app.module.ts` with `enableAppHangTracking` set to `false` to prevent app hanging timeout errors from being reported.
2. Observe the reported errors on the Sentry dashboard.

### Expected Result

- Sentry should not track app hanging errors when enableAppHangTracking is disabled.

### Actual Result

- App hanging timeout errors are still being reported to the dashboard.

### Solution brainstorm

Upon inspecting the TypeScript definition file `option.d.ts`, it appears that the enableAppHangTracking option is missing from BaseCapacitorOptions. 

### Notes

Originally reported on [Discord](https://discord.com/channels/621778831602221064/1211958154984820776/1217704088867966976)

lucas-zimerman:
This feature allow users to disable or increase the time of the app hang events.

Without it, Sentry will create events similar to the following one when the app hangs the UI:
https://sentry-sdks.sentry.io/issues/5080073190/events/018e7e8767624078b7fe1422b415231d

by disabling enableAppHang, Sentry will no longer create the hang events.

Close #596

### DIFF
--- a/src/options.ts
+++ b/src/options.ts
@@ -57,7 +57,29 @@ export interface BaseCapacitorOptions{
    *
    * @default false
    */
+
   enableCaptureFailedRequests?: boolean;
+  /**
+   * When enabled, the SDK tracks when the application stops responding for a specific amount of
+   * time defined by the `appHangTimeoutInterval` option.
+   *
+   * iOS only
+   *
+   * @default true
+   */
+  enableAppHangTracking?: boolean;
+
+  /**
+   * The minimum amount of time an app should be unresponsive to be classified as an App Hanging.
+   * The actual amount may be a little longer.
+   * Avoid using values lower than 100ms, which may cause a lot of app hangs events being transmitted.
+   * Value should be in seconds.
+   *
+   * iOS only
+   *
+   * @default 2
+   */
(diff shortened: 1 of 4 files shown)


src/options.ts
```diff
@@ -57,7 +57,29 @@ export interface BaseCapacitorOptions{
    *
    * @default false
    */
+
```
krystofwoldrich, line 60:
```suggestion
```

src/options.ts
```diff
@@ -58,6 +58,27 @@ export interface BaseCapacitorOptions{
    * @default false
    */
   enableCaptureFailedRequests?: boolean;
+  /**
```
krystofwoldrich, line 61:
```suggestion

  /**
```

PR #605, opened 2024-03-20
Bump: Downgrade Sentry cocoa to 8.21.0

lucas-zimerman:
Version 8.22.2/8.22.1/8.22.0 aren't working with HybridSDKs at the moment and a fix is expected soon, in the meantime we'll bump the SDK to a lower version, not being a downgrade to the users since the latest released version had a lower version for Sentry Cocoa.

This PR also adds an extra script for updating the pod files from the Samples.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,9 +8,9 @@
 
 ### Dependencies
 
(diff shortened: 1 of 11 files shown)


krystofwoldrich (approved):
LGTM 🚀 Thank you

PR #629, opened 2024-04-17
Fix: Breadcrumbs added on forked context are now captured

### LINKED ISSUE #263: Refactor Breadcrumb filters once (team-mobile/issues/11) is completed
Refactor the PR https://github.com/getsentry/sentry-capacitor/pull/254 to not remove the breadcrumbs once https://github.com/getsentry/team-mobile/issues/11 is completed

lucas-zimerman:
This PR brings more native context data to be synced with captured events and also allows hint breadcrumbs to be properly registered, for example:
```typescript
    Sentry.captureException(new Error(`${Date.now()}: a test error occurred`), (context) => {
      return context.addBreadcrumb({ message: 'test' });
    });
```
will now be registered on the event, including the native events:
image

Additional notes on the PR:
- No Required changes were needed to support it on iOS.
- Android had an required refactor in order to capture Envelopes no longer by files.
- some iOS specific properties are now working with both Android and iOS
- Improved the device context sync on both Android and iOS.

## 🛑  Blocked by
- [x] https://github.com/getsentry/sentry-java/issues/3470
- [x] https://github.com/getsentry/sentry-cocoa/issues/4043

Closes #263


src/integrations/devicecontext.ts
```diff
@@ -19,29 +20,119 @@ export class DeviceContext implements Integration {
...
   }
+
+  /**
+ * Merges two groups of ordered breadcrumbs and removes any duplication that may
+ * happen between them.
+ * @param jsList The first group of breadcrumbs from the JavaScript layer.
+ * @param nativeList The second group of breadcrumbs from the native layer.
+ * @returns An array of unique breadcrumbs merged from both lists.
+   */
+  private _mergeUniqueBreadcrumbs(jsList: Breadcrumb[], nativeList: Breadcrumb[]): Breadcrumb[] {
```
krystofwoldrich, line 102: Thank you for adding test for this method, it makes it super easy to understand how it would behave.

m: I'm not sure if we can be sure that the breadcrumbs are in the same order in both arrays.

If I'm looking at the correctly in case they would not be in the same order the duplicates would not be removed. Creating possible double the size.

We should also respect the `maxBreadcrumbs` size.

test/integrations/devicecontext.test.ts
```diff
@@ -0,0 +1,318 @@
...
+          resolve(processedEvent);
+        } catch (e) {
+          reject(e);
+        }
+      }, mockGetCurrentHub);
+    });
+  }
+});
+
+describe('Device Context Breadcrumb filter', () => {
```
krystofwoldrich, line 235: I think test like this would not pass, but from the code I understand that the breadcrumbs should always be ordered the same in both arrays?

```js
const jsList = [
  { timestamp: 2, message: 'new js' },
  { timestamp: 1, message: 'new js' },
  { timestamp: 3, message: 'new js' },
  { timestamp: 4, message: 'new js' },
] as Breadcrumb[];
const nativeList = [
  { timestamp: 2, message: 'new native' },
  { timestamp: 1, message: 'new js' },
  { timestamp: 2, message: 'new js' },
  { timestamp: 3, message: 'new native' },
  { timestamp: 4, message: 'new native' },
] as Breadcrumb[];
```

krystofwoldrich, line 235: We should add test for user set timeout. Something like this:

```js
const jsList = [
  { timestamp: 2, message: 'new js' },
  { timestamp: 1, message: 'new js' },
] as Breadcrumb[];
const nativeList = [
  { timestamp: 2, message: 'new native' },
  { timestamp: 2, message: 'new js' },
  { timestamp: 1, message: 'new js' },
  { timestamp: 3, message: 'new native' },
] as Breadcrumb[];
const expected = [
  { timestamp: 2, message: 'new native' },
  { timestamp: 2, message: 'new js' },
  { timestamp: 1, message: 'new js' },
  { timestamp: 3, message: 'new native' },
] as Breadcrumb[];
```

src/wrapper.ts
```diff
@@ -268,9 +266,6 @@ export const NATIVE = {
     if (!this.enableNative) {
       return;
     }
-    if (!this.isNativeClientAvailable()) {
-      throw this._NativeClientError;
-    }
```
krystofwoldrich, line 273: Why is this removed?

src/wrapper.ts
```diff
@@ -316,22 +310,9 @@ export const NATIVE = {
           // @ts-ignore Android still uses the old message object, without this the serialization of events will break.
           event.message = { message: event.message };
```
krystofwoldrich, line 311: This should still be execute only for Android.

src/wrapper.ts
lucas-zimerman: fixed

src/integrations/devicecontext.ts
lucas-zimerman: I have added tests for the maxBreadcrumb and also for out of order breadcrumbs

android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
```diff
@@ -12,25 +12,33 @@
...
+import java.util.Date;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
+<<<<<<< HEAD
+import java.util.Map;
+=======
 import java.util.Set;
 import java.util.UUID;
+>>>>>>> cf9f7218245fecae3a21f2e6612b0931af53939a
```
krystofwoldrich, line 41: Unresolved merge conflict

```suggestion
import java.util.Map;
import java.util.Set;
import java.util.UUID;
```

android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
```diff
@@ -226,31 +264,14 @@ public void captureEnvelope(PluginCall call) {
...
-            return;
-        }
-        call.resolve();
+            InternalSentrySdk.captureEnvelope(bytes);
+            call.resolve();
+          }
+          catch (Throwable e) {
+            final String errorMessage ="Error while capturing envelope";
+            logger.log(Level.WARNING, errorMessage);
+            call.reject(errorMessage);
```
krystofwoldrich, line 273: I see previously this was `call.reject(String.valueOf(e));`? Does this change the error in any way?

android/src/main/java/io/sentry/capacitor/CapSentryMapConverter.java
```diff
@@ -0,0 +1,137 @@
...
+import io.sentry.ILogger;
+import io.sentry.SentryLevel;
+import io.sentry.android.core.AndroidLogger;
+
+public class CapSentryMapConverter {
+  public static final String NAME = "CapSentry.MapConverter";
+
+  private static final ILogger logger = new AndroidLogger(NAME);
+
+  public static Object convertToWritable(Object serialized) {
```
krystofwoldrich, line 22: Can this return `JSObject` directly?

android/src/main/java/io/sentry/capacitor/CapSentryMapConverter.java
```diff
@@ -0,0 +1,137 @@
...
+import java.math.BigDecimal;
+import java.math.BigInteger;
+import java.util.List;
+import java.util.Map;
+
+import io.sentry.ILogger;
+import io.sentry.SentryLevel;
+import io.sentry.android.core.AndroidLogger;
+
+public class CapSentryMapConverter {
```
krystofwoldrich, line 17: Can we add unit tests for this class?

src/integrations/devicecontext.ts
```diff
@@ -19,29 +20,127 @@ export class DeviceContext implements Integration {
...
+    // Add remaining breadcrumbs from the JavaScript and Native list if space allows.
+    while (jsIndex < orderedJsList.length  && combinedList.length < maxBreadcrumbs) {
+      combinedList.push(orderedJsList[jsIndex++]);
+    }
+
+    while (natIndex < orderedNativeList.length && combinedList.length < maxBreadcrumbs) {
+      combinedList.push(orderedNativeList[natIndex++]);
+    }
+    return combinedList;
+  }
```
krystofwoldrich, line 145: The function works, but I think we should consider marking the JS crumbs during the creation on native layers, as then we can also filter then on the native layers and only return the native crumbs.

The native solution would be faster and required less memory.

We could do `JSBreadcrumb extends Breadcrumb` and then filter `breadcrumb instanceof JSBreadcrumb`.

romtsn: yeah, let's please not use `unknown` - it exists to work-around JVM's static type system exclusively. A new field is the cleanest way imo, but we could also open the `Breadcrumb` class for extension, should be not a problem

lucas-zimerman: I also like the idea of breadcrumb.origin, it could also help in the future to know if a breadcrumb was originated from the JavaScript layer or native

antonis: > I also like the idea of breadcrumb.origin, it could also help in the future to know if a breadcrumb was originated from the JavaScript layer or native

Hey @lucas-zimerman 👋 
I've [drafted an iOS PR](https://github.com/getsentry/sentry-cocoa/pull/4358/files) and was wondering if you have any early feedback on it. Does the approach chosen unblock this PR?

lucas-zimerman: It surelly does thank you!

PR #645, opened 2024-05-11
Bump: Sentry JavaScript 7.114.0

lucas-zimerman:
This PR bumps Sentry from version 93 up to 114.0.
No break changes were seen on both Windows and MacOS


.eslintrc.js
```diff
@@ -48,5 +48,6 @@ module.exports = {
     '@sentry-internal/sdk/no-optional-chaining': 'off',
     '@sentry-internal/sdk/no-nullish-coalescing': 'off',
     '@sentry-internal/sdk/no-class-field-initializers': 'off',
+    'deprecation/deprecation': 'off',
```
krystofwoldrich, line 51: Are there many deprecation warnings? (More than ~20?)

If so this fine, but if it's not too many I would prefer disabling it locally to make it obvious which APIs will need to be updated in `sentry-capacitor` and the examples.

But let's keep it for next PR.

krystofwoldrich (approved):
LGTM 🚀

Issue #630, opened 2024-04-18
feat: Update to capacitor 6


EinfachHans:
As Capacitor 6 is officially released, it would be awesome if you can also release a V6 including support for it.

The migration is quite easy via their migration: https://capacitorjs.com/docs/updating/plugins/6-0

EinfachHans:
@lucas-zimerman This is the last plugin that blocks me from updating to v6, anything to share already?

I can also assist with this, but because you don't stick to the capacitor versioning and support multiple capacitor versions at the same time i have problems to do so.

lucas-zimerman:
Hi @EinfachHans and thanks for creating the issue! Support for Capacitor 6 is planned for the next release

EinfachHans:
not stale

RobSchilderr:
What's the status of this issue?

lucas-zimerman:
The support for capacitor 6 was released, if you have more questions, feel free to create a new issue

PR #651, opened 2024-05-15
Feat: Support Capacitor 6

lucas-zimerman:
Fixes: #630

This PR brings support to Capacitor while also adding some quality of life improvements. During the review, the ionic-angular-v2 was removed due to being a too old dependency of Capacitor that no longer works on the latest Android Studio, I opted to remove it since we already will cover a wide range of versions (3.0 up to 6.0). Also ionic-angular was renamed to ionic-angular-v4 to match the pattern where the last number indicates the capacitor version used by the sample.

Minor changes:

Samples:
- Added Capacitor 6 and 3 Samples.
- Removed Capcitor 2 Sample.
- Added script to skip pod install on Windows.

SDK:
- Add Required changes to support Capacitor 6.
- Code cleanup on gradle script in order to reduce dependencies and make it more simple to maintain.
-


android/build.gradle
```diff
@@ -1,60 +1,58 @@
...
+    return 6
+  }
+  else if (compileVersion == 33) {
+    return 5
+  }
+  else if (compileVersion == 32) {
+    return 4
+  }
+  else {
+    throw new GradleException("The version of Capacitor is not supported by Sentry, CODE: " + compileVersion)
```
krystofwoldrich, line 23: This seems like a breaking change.

Did the SDK build with Capacitor 3 before?

example/ionic-angular-v3/scripts/skip-pod-windows.js
krystofwoldrich, line 1: This script is not used. Probably duplicate from the top level scripts.

android/build.gradle
krystofwoldrich: Or does the case for 4 covers 3 as well?

krystofwoldrich:
The SDK is in v0.X.X so it's okay to drop some very old capacitor version, but we should mentioned in Changelog.

I'm not sure if we dropped any, from the code it looks like maybe 2 and 3?

android/build.gradle
lucas-zimerman: Nice catch, I'll fix the case for Capacitor 3

package.json
```diff
@@ -70,9 +70,9 @@
     "@sentry/utils": "7.114.0"
   },
   "devDependencies": {
-    "@capacitor/android": "^3.0.1 || ^4.0.0 || ^5.0.0",
-    "@capacitor/core": "^3.0.1 || ^4.0.0 || ^5.0.0",
-    "@capacitor/ios": "^3.0.1 || ^4.0.0 || ^5.0.0",
+    "@capacitor/android": "^3.0.1 || ^4.0.0 || ^5.0.0 || ^6.0.0",
+    "@capacitor/core": "^3.0.1 || ^4.0.0 || ^5.0.0 || ^6.0.0",
+    "@capacitor/ios": "^3.0.1 || ^4.0.0 || ^5.0.0 || ^6.0.0",
```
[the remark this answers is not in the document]
lucas-zimerman: it's valid

jcesarmobile:
Capacitor 4 and older are out of support, I don’t think you should support those versions anymore

https://capacitorjs.com/docs/main/reference/support-policy

android/build.gradle
```diff
@@ -1,60 +1,58 @@
...
-    if (rootProject.project(':capacitor-android').android.compileOptions.sourceCompatibility >= JavaVersion.VERSION_17) {
-      println "Sentry Capacitor: Applying dependencies for Capacitor 5 or higher."
-      classpath 'com.android.tools.build:gradle:8.0.0'
-      classpath 'com.google.gms:google-services:4.3.15'
-    }
-    else {
-      println "Sentry Capacitor: Applying dependencies for Capacitor 4."
-      classpath 'com.android.tools.build:gradle:7.2.1'
-      classpath 'com.google.gms:google-services:4.3.13'
-    }
```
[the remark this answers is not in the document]
lucas-zimerman: the SDK doesn't depend on google services and the gradle version is going to be used from the parent project. Making this code unnecessary and harder to maintain once new versions arrive.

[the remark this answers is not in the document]
krystofwoldrich: Awesome, thanks for the explanation.

krystofwoldrich (approved):
Looks good, thank you! 🚀

Issue #647, opened 2024-05-12
Calling setTag with undefined value causes the SDK to throw


vincentbrison:
### Environment

sentry/capacitor 0.17.0, on iOS

### Steps to Reproduce

Calling `Sentry.setTag("someKey", undefined)` causes the error `{"message":"Error deserializing tag","errorMessage":"Error deserializing tag"}`, which itself is reported to Sentry. Maybe the native part is not handling `undefined` correctly?

### Expected Result

It should not report any error, as calling `setTag` with an undefined value should only unset the tag if any.

### Actual Result

Throwing the error `{"message":"Error deserializing tag","errorMessage":"Error deserializing tag"}`

lucas-zimerman:
Hi and thank you for reporing this issue!

When setting undefined, are you trying to remove the tag "someKey"?

vincentbrison:
Not necessarily, our app is setting a wide range of tags and depending the usage, certain tags might be undefined. So our current work around is to not set undefined tags, but looking at the doc it still looks like an issue.

PR #656, opened 2024-05-24
Fix: Accept undefined as value for tags.

lucas-zimerman:
the behaviour on JavaScript with undefined value is to remove the tag, this wasn't respected on the native SDKs, this PR ensures that the behaviour form JavaScript also happens on the Native layer, by removing the tag.

Fixes: #647

### DIFF
--- a/android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
+++ b/android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
@@ -347,14 +347,23 @@ public class SentryCapacitor extends Plugin {
 
     @PluginMethod
     public void setTag(PluginCall call) {
-        if (call.getData().has("key") && call.getData().has("value")) {
+        if (call.getData().has("key")) {
+          if (call.getData().has("value")) {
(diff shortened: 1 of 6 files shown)


krystofwoldrich (approved):
LGTM 🚀 

I guess we can't add easy test to check that `Sentry.setTag('tagKey', undefined)` is translated by Capacitor to `call.hasKey('value')===false`.

android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
krystofwoldrich, line 1: I think using the if as guards to leave early like in the Swift implementation is easier to leave.

I leave it up to you if you would like to change it.

krystofwoldrich:
@lucas-zimerman Could you open the same fix for React Native?

android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
lucas-zimerman: sure thing I'll adjust it

PR #664, opened 2024-06-04
Feat: Add close function to the main SDK.

### LINKED ISSUE #663: getCurrentHub: Incorrect deprecation notice?
Hey all!

I'm currently trying to close the Sentry connection on runtime whenever we detect an app update is available and the user has been prompted. As I'm lead to believe the current way to do so is as below, however, this prompts a deprecation notice to use `Sentry.getClient()` instead. This method does not exist, thus it seems this deprecation notice is incorrect.

```js
Sentry.getCurrentHub().getClient().close(0); // deprecated, but works
Sentry.getClient().close(0); // Sentry.getClient is not a function
```

 In contrast it does seem to work correctly when trying to close a Vue-specific instance of Sentry. In other words, this following prompts (correctly) to use the alternative:

```js
SentryVue.getCurrentHub().getClient().close(0); // deprecated, but works
SentryVue.getClient().close(0); // works just fine
```

While having to go through this many hoops to simply disable Sentry seems like a drag, that is a separate issue, though I would like to know whether this is correct, or whether only `Sentry` or `SentryVue` needs to be closed.

I'm currently using the following Sentry-relevant packages:

```json
"@sentry/capacitor": "0.18.0",
"@sentry/vite-plugin": "2.18.0",
"@sentry/vue": "7.114.0",
```

lucas-zimerman:
This PR exposes the close function that closes the client and the native layer.

Minor changes:

# Sample
* Removed deprecated flags from the latest samples.
* Added a button to call the close function.
Closes #663

### DIFF
--- a/src/wrapper.ts
+++ b/src/wrapper.ts
@@ -279,14 +279,19 @@ export const NATIVE = {
   /**
    * Closes the Native Layer SDK
    */
-  closeNativeSdk(): Promise<void> {
-    if (!this.enableNative || !this.isNativeClientAvailable()) {
+   closeNativeSdk(): Promise<void> {
+    if (!this.enableNative) {
+      // Only log and avoid rejecting.
+      logger.debug(this._DisabledNativeError);
       return Promise.resolve();
     }
+    if (!this.isNativeClientAvailable()) {
+      throw this._NativeClientError;
+    }
 
-    return SentryCapacitor.closeNativeSdk().then(() => {
-      this.enableNative = false;
-    });
+     return SentryCapacitor.closeNativeSdk().then(() => {
+       this.enableNative = false;
+     });
   },
 
 
--- a/src/wrapper.ts
+++ b/src/wrapper.ts
@@ -275,6 +275,21 @@ export const NATIVE = {
     SentryCapacitor.clearBreadcrumbs();
   },
 
+
+  /**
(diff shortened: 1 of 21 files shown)


test/wrapper.test.ts
```diff
@@ -625,4 +626,14 @@ describe('Tests Native Wrapper', () => {
       expect(SentryCapacitor.fetchNativeDeviceContexts).not.toBeCalled();
     });
   });
+
+  describe('closeNativeSdk', () => {
```
krystofwoldrich, line 630: The current test looks good. Let's add one more for two calls after each other, to ensure this won't throw in future.

src/wrapper.ts
```diff
@@ -275,6 +275,21 @@ export const NATIVE = {
...
   },
 
+
+  /**
+   * Closes the Native Layer SDK
+   */
+  closeNativeSdk(): Promise<void> {
+    if (!this.enableNative || !this.isNativeClientAvailable()) {
+      return Promise.resolve();
+    }
```
krystofwoldrich, line 285: Let's split this into two if's and add logger to log out the information why close on native was not called.

krystofwoldrich:
Thank you looks good, I just have two small nits.

krystofwoldrich (approved):
LG 🚀 Thank you!

PR #727, opened 2024-09-09
fix:remove old dependencies

lucas-zimerman:
This PR removes version check for dependencies that no longer exists on V8

#skip-changelog


krystofwoldrich (approved):
LGTM 🚀

PR #728, opened 2024-09-09
fix: wrong pod lock files version

lucas-zimerman:
Fix pod lock with the wrong version, should only impact CI builds.

#skip-changelog.


krystofwoldrich (approved):
LGTM 🚀

PR #729, opened 2024-09-09
Update changelog

lucas-zimerman:
Since 0.19 was released outside of main, we don't have the changelog of it.
0.19 contains fixes that were included on the latest beta so I am only bumping the changelog

#skip-changelog


krystofwoldrich (approved):
LGTM 🚀  

We could also open PR for the release branch and merge it to main. But this works too.

PR #766, opened 2024-10-29
Fix: normalize should use Options limits or default

lucas-zimerman:
This change allows Breadcrumbs to respect the defined limits of `normalizeDepth` and `normalizeMaxBreadth`.
Before that, it would use the limits of 100 for `normalizeDepth` and infinite for `normalizeMaxBreadth`

The PR also updates GitHub Checks that were using old versions, which were blocking the build to pass.


.github/workflows/buildandtest.yml
krystofwoldrich, line 1: l: It's small PR, so it's fine here.

But it would be more clear to split the GH changes into separate PR.

PR #776, opened 2024-11-11
 fix(core): Export missing span utils from Sentry Core 

lucas-zimerman:
This PR adds missing span utils export from @sentry/core.


krystofwoldrich (approved):
LGTM 🚀

PR #777, opened 2024-11-11
Ref: use --sig parameter when bumping samples

lucas-zimerman:
Ionic usually gets the wrong of the SDK, despite the right one being installed, adding the --sig parameter to publish will add the hash of the SDK and will avoid conflicts with wrong versions being tested with the sample.

More information: https://github.com/wclr/yalc?tab=readme-ov-file#publish

#skip-changelog


krystofwoldrich (approved):
Nice, I didn't know this option before!

PR #778, opened 2024-11-11
internal: Enable corepack

lucas-zimerman:
Since other SDKs are using corepack and the change was simple here, I thought on enabling it here so it avoids corepack trying to install the incorrect version of yarn every time I work on Capacitor with corepack enabled.

#skip-changelog.


krystofwoldrich (approved):
LGTM 🚀

PR #795, opened 2024-11-26
feat/WIP:   e2e sibling test

lucas-zimerman:
This PR introduces new tests to evaluate how package managers behave when installing or updating sibling SDKs alongside Sentry Capacitor.
# Tests Included:

1. Updating Existing Installations
   * Run yarn install/npm install when both Sentry Capacitor and a sibling SDK are already installed to update the libraries.

2. Adding a Sibling SDK
     *  Install a sibling SDK using yarn add ... or npm install ... when Sentry Capacitor is already installed.

3. Fresh Installations
     *  Install both Sentry Capacitor and a sibling SDK when neither is currently installed.

# Validation:
For each test involving sibling SDKs:

- Verify success when the correct sibling versions are installed.
- Ensure failures occur for incompatible versions.

# Coverage:

This PR will cover npm, Yarn Classic, and Yarn v3. By adding these tests, we significantly enhance the coverage for installation and update scenarios involving Sentry Capacitor, ensuring better reliability and user experience.

#skip-changelog


scripts/e2e-test.siblings.sh
```diff
@@ -0,0 +1,23 @@
...
+
+# Navigate to siblingsTests and publish without modifying package.json
+cd e2e-test/scripts/siblingsTests/
+
+rm -r .yalc
+rm -r node_modules
+
+cp package.json package.json.bak
+
+yalc add @sentry/capacitor
```
lucas-zimerman, line 14: We create this folder because it loads faster than pointing yarn to install the root project (10 minutes vs 5 seconds) on the sibling tests

krystofwoldrich: By this folder, do you mean `e2e-test/scripts/siblingsTests`?

test/scripts/check-siblingsHelper.ts
```diff
@@ -0,0 +1,221 @@
...
+  fs.writeFileSync(newPackagePath, JSON.stringify(mergedPackageJson, null, 2), 'utf8');
+}
+
+export function InstallSDK(pkgMnger: string, args: ReadonlyArray<string>, rootPath: string): ReturnType<typeof spawnSync>
+{
+  const outputStream = fs.openSync(path.join(rootPath, installLogFilename), 'w');
+
+  const nodePatch = pkgMnger === 'npm' ? {
+    npm_package_json: path.join(rootPath, 'node_modules', '@sentry', 'capacitor', 'package.json')
+  } : {};
```
krystofwoldrich, line 100: Is this still used anywhere? In the check file I've noticed the `npm_package_json` was deleted.

test/scripts/check-siblingsHelper.ts
```diff
@@ -0,0 +1,221 @@
...
+
+export function GetE2ELogs(testPath: string): string[] {
+  const data = fs.readFileSync(path.join(testPath, installLogFilename), 'utf8');
+  return FilterLogs(data, true);
+}
+
+export function GetLogs(testPath: string): string[] {
+  const data = fs.readFileSync(path.join(testPath, installLogFilename), 'utf8');
+  return FilterLogs(data, false);
+}
```
krystofwoldrich, line 221: Do we need the filtering logic? If I understand it correctly we anyway use `contains` to check if the message is present in the logs, so I would expect it to pass also without the filtering.

krystofwoldrich:
Overall the tests run and look correct, but I'm surprised by the complexity and number of helpers needed.

---

It's not a blocker, but have you considered preparing complete example project directories (with complete package.json) in which only `manage install` would need to be executed? Instead of creating the directories dynamically?

test/scripts/check-siblingsHelper.ts
lucas-zimerman: No, good catch! I removed it on https://github.com/getsentry/sentry-capacitor/pull/795/commits/ab94537e039271067fca5edbc5d21d99aa0bf6bb

krystofwoldrich:
The tests look good.

But the CI fails due to deprecated `actions/cache@v4.1.2`.

PR #803, opened 2024-12-09
Bugfix: tracePropationTargets option ignored in CapacitorHTTP methods

wif70068:
This update addresses an issue where the tracePropagationTargets option was being ignored in various CapacitorHTTP methods.


wif70068:
keep alive

lucas-zimerman (approved):

lucas-zimerman:
Sorry we will have to rollback this PR due to extra complexity that will be required in order to implement it.

But I have opened an issue here to keep track of this feature https://github.com/getsentry/sentry-capacitor/issues/820

wif70068:
It works for my specific use case. Are you saying this PR breaks other functionality?

Issue #830, opened 2025-01-30
Support Capacitor 7


niconaso:
[Capacitor 7 received its first stable release last week.](https://github.com/ionic-team/capacitor/releases/tag/7.0.0)

It would be great for this plugin to add cap 7 support.

lucas-zimerman:
Thank you for opening this issue! The support for Capacitor 7 will start the development shortly, if there aren't any major break changes, expect a new release next week

PR #831, opened 2025-01-31
Feat: Capacitor v7 support

lucas-zimerman:
Depends on: #832
Fixes: #830

This PR adds support for Capacitor V7 to the SDK.
No break changes were introduced.


SentryCapacitor.podspec
```diff
@@ -17,15 +17,19 @@ Pod::Spec.new do |s|
 
   if File.exist?('../../@capacitor/core/package.json') == false
     # If Capacitor was not found (could happen when using Yarn PNP), fallback to the
-    # required minimum version of Capacitor 4.
-    miniOSVersion = '13.0'
+    # required minimum version of Capacitor 7.
+    miniOSVersion = '14.0'
```
krystofwoldrich, line 21: This could be potentially breaking for some users, but since it's the fallback I think it's okay.

SentryCapacitor.podspec
```diff
@@ -17,15 +17,19 @@ Pod::Spec.new do |s|
...
-      miniOSVersion = '12.0'
+    majorVersion = capacitorVersion.split('.').first.to_i
+
+    if majorVersion <= 3
+      miniOSVersion = '12.0' # Capacitor 3, 2.
+    elsif majorVersion < 7
+      miniOSVersion = '13.0' # Capacitor 6 or older.
     else
-      miniOSVersion = '13.0' # Required for Capacitor 4 and newer.
+      miniOSVersion = '14.0' # Capacitor 7 or higher.
```
krystofwoldrich, line 32: Does the Capacitor V7 fail to build without the matching minimum target?

I would have expected that a library could specify an older target than the framework.

krystofwoldrich (approved):
LGTM 🚀

SentryCapacitor.podspec
lucas-zimerman: it did for me, the sdk was set with min iOS version 13 and it was complaining that the project required version 14.

lucas-zimerman:
Dependency Review failed with issues in regard to the new sample, those issues will not impact the user code so I will only take actions about it after the PR

PR #840, opened 2025-02-11
feat: Sentry Javascript V9

### LINKED ISSUE #819: Explicity set user.ip_address: {{auto}}
See https://github.com/getsentry/team-sdks/issues/118

Issue will probably be done once we bump a JS SDK version with https://github.com/getsentry/sentry-javascript/pull/15008 is released

### LINKED ISSUE #837: Implement Sampling Seed Propagation
### Description

See https://github.com/getsentry/projects/issues/316

To increase the chance of capturing complete traces when users return a new sample rate tracesSampler in backend services, we propagate the random value used by the SDK for computing the sampling decision instead of creating a new random value in every service. Therefore, across a trace every SDK uses the same random value.

[Develop: **Spec**](https://develop.sentry.dev/sdk/telemetry/traces/#propagated-random-value)

Similar to https://github.com/getsentry/sentry-react-native/issues/4425 we most likely need to only bump the JS SDK.

### See also

Sample implemeantion in PHP https://github.com/getsentry/sentry-php/pull/1793.

lucas-zimerman:
The V9 change introduces break changes, but not as heavy as V8, making it without any break changes to some users.

- Main break changes

* Typescript bumped to V5
* SDK initialization with VUE 
* `sentry/utils` no longer exists, main features were moved to `sentry/core`
* JavaScript SDK now accepts multiple versions in parallel (not sure how it will behave)

Fixes #819 
Fixes #837


src/transports/encodePolyfill.ts
```diff
@@ -2,13 +2,19 @@ import { CAP_GLOBAL_OBJ } from '../utils/webViewUrl';
...
-  if (!CAP_GLOBAL_OBJ.__SENTRY__) {
-    (CAP_GLOBAL_OBJ.__SENTRY__ as Partial<
+  const globalCarriers = CAP_GLOBAL_OBJ.__SENTRY__;
+
+  if (!globalCarriers) {
+    (globalCarriers as Partial<
       (typeof CAP_GLOBAL_OBJ)['__SENTRY__']
     >) = {};
   }
+  const capacitorSiblingCarrier = globalCarriers?.version && globalCarriers['9.0.0'];
```
lucas-zimerman, line 12: Q: should we use the version used by the SDK or the newest carrier?

lucas-zimerman: used similar implementation as react native, load version using SDK_VERSION from @sentry/core

CHANGELOG.md
```diff
@@ -2,10 +2,19 @@
 
 <!-- prettier-ignore-start -->
 > [!IMPORTANT]
-> If you are upgrading to the `1.x` versions of the SDK from `0.x`, make sure you follow our
+> If you are upgrading to the `2.x` versions of the SDK from `1.x`, make sure you follow our
```
lucas-zimerman, line 5:
```suggestion
> If you are upgrading to the `2.x` versions of the SDK from `1.x` or lower, make sure you follow our
```

package.json
```diff
@@ -60,23 +60,22 @@
...
-    "@sentry-internal/eslint-plugin-sdk": "8.42.0",
-    "@sentry-internal/typescript": "8.42.0",
+    "@sentry-internal/eslint-config-sdk": "9.0.0",
+    "@sentry-internal/eslint-plugin-sdk": "9.0.0",
+    "@sentry-internal/typescript": "9.0.0",
     "@sentry/wizard": "3.34.2",
     "@types/jest": "^29.5.3",
     "concurrently": "^8.2.2",
-    "eslint": "^7.13.0",
+    "eslint": "^7.6.0",
```
krystofwoldrich, line 78: Did you mean to downgrade eslint here?

src/transports/encodePolyfill.ts
```diff
@@ -1,14 +1,21 @@
...
+    (globalCarriers as Partial<
       (typeof CAP_GLOBAL_OBJ)['__SENTRY__']
     >) = {};
   }
+  const capacitorSiblingCarrier = CAP_GLOBAL_OBJ.__SENTRY__?.[SDK_VERSION];
 
-  CAP_GLOBAL_OBJ.__SENTRY__.encodePolyfill = encodePolyfill;
+  if (capacitorSiblingCarrier) {
+    capacitorSiblingCarrier.encodePolyfill = encodePolyfill;
+  }
```
krystofwoldrich, line 18: Because this is called before the browserInit the `capacitorSiblingCarrier ` won't be set and thus the `encodePolyfill` won't be set.

JS has a helper function https://github.com/getsentry/sentry-javascript/blob/fc0af3892a2704326d2338c05d3f049176d9358d/packages/core/src/carrier.ts#L42 which ensures the version carrier is always ready.

If it's not ready we should log an error because that's unexpected.

CHANGELOG.md
```diff
@@ -2,10 +2,19 @@
...
 
+## Unreleased
+
+### Dependencies
+
+- Bump JavaScript SDKs from v8.42.1 to v9.0.0 ([#840](https://github.com/getsentry/sentry-capacitor/pull/840))
+
+  - [changelog](https://github.com/getsentry/sentry-javascript/blob/9.0.0/CHANGELOG.md#900)
+  - [diff](https://github.com/getsentry/sentry-javascript/compare/8.42.0...9.0.0)
+
```
krystofwoldrich, line 17: We could also add the migration guide to the list.

````suggestion
  - [migration guide](https://docs.sentry.io/platforms/javascript/migration/v8-to-v9/)

````

Comment (krystofwoldrich) [reply CHANGELOG.md:17]: Actually, we can update the changelog when https://github.com/getsentry/sentry-docs/pull/13022 is merged.

**Comment by krystofwoldrich** [comment]
Looks good, I have only one major comment about the global carrier.

Before Cap SDK v2 release I think we should follow this up with a PR updating the SDK to the latest JS SDK v9.

File: src/transports/encodePolyfill.ts (hunk 4 above)
Comment (lucas-zimerman) [reply src/transports/encodePolyfill.ts:18]: Sure thing I will use it, although I am not sure if it will always be ready according to the function comment  `FIXME: This function is problematic`

Comment (krystofwoldrich) [reply src/transports/encodePolyfill.ts:18]: The reason for the `FIXME: This function is problematic` is that the type still returns the carrier marked as optional, even tho it's created when missing.

But it's used in the JS SDK, so I think we should use it too or ensure in a different way that the carrier is created.

File: package.json (hunk 3 above)
Comment (lucas-zimerman) [reply package.json]: Nice catch! fixed on https://github.com/getsentry/sentry-capacitor/pull/840/commits/da2eb29901775772251ee0abed1e8476ba429ad4

File: src/transports/encodePolyfill.ts (hunk 4 above)
Comment (lucas-zimerman) [reply src/transports/encodePolyfill.ts:17]: I see, using main carrier on https://github.com/getsentry/sentry-capacitor/pull/840/commits/2e8c915868248d9d56426313d69a7a7a0368548d

**Comment by lucas-zimerman** [comment]
> Looks good, I have only one major comment about the global carrier.
> 
> Before Cap SDK v2 release I think we should follow this up with a PR updating the SDK to the latest JS SDK v9.

Follow up PR Here: https://github.com/getsentry/sentry-capacitor/pull/872

File: CHANGELOG.md [hunk 6]
```diff
@@ -2,14 +2,19 @@
...
 ## Unreleased
 
 ### Dependencies
 
+- Bump JavaScript SDKs from v8.55.0 to v9.0.0 ([#840](https://github.com/getsentry/sentry-capacitor/pull/840))
+
+  - [changelog](https://github.com/getsentry/sentry-javascript/blob/9.0.0/CHANGELOG.md#900)
+  - [diff](https://github.com/getsentry/sentry-javascript/compare/8.55.0...9.0.0)
+
 - Bump Sentry Javascript to 8.55.0 ([#870](https://github.com/getsentry/sentry-capacitor/pull/870))
```
krystofwoldrich, line 18: Let's release the 8.55.0 bump first, and then the v9.

src/transports/encodePolyfill.ts
```diff
@@ -1,14 +1,20 @@
...
+    (globalCarriers as Partial<
       (typeof CAP_GLOBAL_OBJ)['__SENTRY__']
     >) = {};
   }
+  const capacitorSiblingCarrier = globalCarriers?.[SDK_VERSION];
 
-  CAP_GLOBAL_OBJ.__SENTRY__.encodePolyfill = encodePolyfill;
+  if (capacitorSiblingCarrier) {
+    capacitorSiblingCarrier.encodePolyfill = encodePolyfill;
+  }
```
krystofwoldrich, line 17: We can simplify this by copying what JS is doing.

````suggestion
  // Based on https://github.com/getsentry/sentry-javascript/blob/f0fc41f6166857cd97a695f5cc9a18caf6a0bf43/packages/core/src/carrier.ts#L49
  const carrier = getMainCarrier();
  const __SENTRY__ = (carrier.__SENTRY__ = carrier.__SENTRY__ || {});
  (__SENTRY__[SDK_VERSION] = __SENTRY__[SDK_VERSION] || {}).encodePolyfill = encodePolyfill;
````

CHANGELOG.md
lucas-zimerman: 8.55.0 was already releasem i fixed the merge conflct here https://github.com/getsentry/sentry-capacitor/pull/840/commits/2f7bd2916a7509cb5863bf76af581a6ae54e1c96

krystofwoldrich (approved):
LG 🚀

Issue #908, opened 2025-05-28
cocoa update doesn't update the lock


lucas-zimerman:
```
[!] CocoaPods could not find compatible versions for pod "Sentry/HybridSDK":
  In snapshot (Podfile.lock):
    Sentry/HybridSDK (= 8.50.1)

  In Podfile:
    SentryCapacitor (from `../../node_modules/@sentry/capacitor`) was resolved to 2.0.0-beta.1, which depends on
      Sentry/HybridSDK (= 8.51.1)
```
ATM it only updates the SDK number, but it doesn't update the samples and the lock files.

The updater should also update those files when updating a PR

lucas-zimerman:
With the recent change of https://github.com/getsentry/sentry-capacitor/pull/938 this problem will no longer happen

PR #909, opened 2025-05-28
 Fix pods

lucas-zimerman:
https://github.com/getsentry/sentry-capacitor/pull/895
The cocoa updater is not updating all the fields so CI tests are failing to pass..

This PR updates the pod files manually with the current Sentry Cocoa SDK.

An issue was open to track the main problem of the update cocoa  script https://github.com/getsentry/sentry-capacitor/issues/908

#skip-changelog.

### DIFF
--- a/example/ionic-angular-v6/ios/App/App/capacitor.config.json
+++ b/example/ionic-angular-v6/ios/App/App/capacitor.config.json
@@ -9,6 +9,8 @@
 	"webDir": "www",
 	"packageClassList": [
 		"SentryCapacitor",
-		"SentryCapacitor"
+		"kSentryLevelNone",
+		"kSentryTransactionNameSourceCustom",
+		"SentryRRWebEvent"
 	]
 }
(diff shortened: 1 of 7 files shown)


example/ionic-angular-v6/ios/App/App/capacitor.config.json
```diff
@@ -9,6 +9,8 @@
 	"webDir": "www",
 	"packageClassList": [
 		"SentryCapacitor",
-		"SentryCapacitor"
+		"kSentryLevelNone",
+		"kSentryTransactionNameSourceCustom",
+		"SentryRRWebEvent"
```
antonis, line 14: q: why are those classes added?

lucas-zimerman: That is a good question, it seems to be only added on later versions of capacitor automatically.
I will investigate later what's up with it but so far it doesn't affect the sdk usage.

Issue #854, opened 2025-03-14
# Feature Request: Add Swift Package Manager (SPM) Support for sentry-capacitor


YoannRu:
## Context
Currently, the `sentry-capacitor` package is only available via Cocoapods for iOS integration. To provide greater flexibility for developers and meet the needs of modern projects, it would be beneficial to add support for Swift Package Manager (SPM).

## Current Issue
- **Limited Dependency Management**: Relying exclusively on Cocoapods can be a constraint for projects that prefer or require SPM.
- **Growing Adoption of SPM**: An increasing number of developers and projects are choosing SPM due to its native integration in Xcode and ease of use.

## Proposed Solution
- **SPM Integration**: Add a configuration with a `Package.swift` file that allows `sentry-capacitor` to be integrated directly via SPM.
- **Testing and Validation**: Implement tests to ensure that the SPM integration works correctly and handles all necessary dependencies.
- **Documentation**: Update the documentation with clear instructions for installation and usage of the package with SPM. For more details on SPM integration, please refer to the following guide: [SPM Integration Guide](https://capacitorjs.com/docs/ios/spm).

## Environment
- **Platform**: iOS
- **Current Package Manager**: Cocoapods
- **Proposed Package Manager**: Swift Package Manager (SPM)

Thank you for considering this request. Please feel free to reach out if you have any questions or need further details.

Sincerely,

PR #938, opened 2025-07-17
Migrate Plugin to SPM

jb3rndt:
Apps that use SPM are unable to use this plugin as it does not support SPM. As requested in https://github.com/getsentry/sentry-capacitor/issues/854, this PR converts the package to SPM.

LMK what you think :)

Close https://github.com/getsentry/sentry-capacitor/issues/854


ios/Sources/SentryCapacitorPlugin/SentryCapacitorPlugin.swift
```diff
@@ -1,13 +1,32 @@
 import Foundation
 import Capacitor
 import Sentry
+import Sentry._Hybrid
```
lucas-zimerman, line 4: Q: Why have you added the import to Sentry Hybrid?

lucas-zimerman, line 7: Lets add this option optional since it is only required when users use SPM.

```suggestion

// Keep compatibility with CocoaPods.
#if SWIFT_PACKAGE
import Sentry._Hybrid
#endif
```

jb3rndt: Good point 👍

lucas-zimerman (approved):

PR #946, opened 2025-08-06
Add swift package to cocoa updater

lucas-zimerman:
With recent changes, we now have to update the podfile along with the swift package, this Package fix the updater to update both packages.

#skip-changelog.


antonis (approved):
LGTM!

PR #957, opened 2025-08-13
FIX: missing ip settings from SDK package

lucas-zimerman:
Based on https://github.com/getsentry/sentry-javascript/pull/17364

original context;
```
This PR fixes a long-standing problem in the SDK where it would set incorrect information about when Relay should (not) infer IP addresses from sent envelope requests.

Previously, this was thought to be controlled by setting event.user.ip_address: '{{auto}}'. However, after an incident in Relay, it was determined that this is in fact not a reliably way to control IP inference. Instead, SDKs should set event.sdk.settings.infer_ip: 'auto' | 'never' (see closes https://github.com/getsentry/sentry-javascript/issues/16252).
Unfortunately, this wasn't implemented immediately but is taken care of in this PR.

(FWIW, the only reason why Relay continued to infer IP addresses for the JS SDK was because it is excempt from logic that would infer IP addresses only if user.ip_address was set to '{{auto}}'. This is necessary to backwards compatibility with older SDKs.)

Follow-ups: We likely also need to adjust the logic in Electron and Lynx (at the very least remove setting user.ip_address).

closes https://github.com/getsentry/sentry-javascript/issues/17351
closes https://github.com/getsentry/sentry-javascript/issues/16252
```

Instead of bumping o V10, the fix was patched on the current release.

### DIFF
--- a/example/ionic-angular-v7/ios/App/App/capacitor.config.json
+++ b/example/ionic-angular-v7/ios/App/App/capacitor.config.json
@@ -8,6 +8,9 @@
 	},
 	"webDir": "www",
 	"packageClassList": [
-		"SentryCapacitorPlugin"
+		"SentryRRWebEvent",
+		"kSentryTransactionNameSourceCustom",
+		"kSentryLevelNone",
+		"SentryCapacitor"
 	]
 }
--- a/example/ionic-angular-v7/ios/App/App/capacitor.config.json
+++ b/example/ionic-angular-v7/ios/App/App/capacitor.config.json
@@ -8,9 +8,6 @@
 	},
 	"webDir": "www",
 	"packageClassList": [
-		"SentryRRWebEvent",
-		"kSentryTransactionNameSourceCustom",
-		"kSentryLevelNone",
-		"SentryCapacitor"
+		"SentryCapacitorPlugin"
 	]
 }
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
     "type": "git",
     "url": "https://github.com/getsentry/sentry-capacitor"
   },
-  "version": "2.1.3",
+  "version": "2.1.0",
   "description": "Official Sentry SDK for Capacitor",
   "types": "dist/esm/index.d.ts",
   "main": "dist/build/index.js",
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
     "type": "git",
     "url": "https://github.com/getsentry/sentry-capacitor"
   },
-  "version": "2.1.0",
+  "version": "2.1.3",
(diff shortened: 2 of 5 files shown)


lucas-zimerman:
auto set https://us.sentry.io/api/0/projects/sentry-sdks/sentry-capacitor/events/dbac1fe89e3b46d6a4b49cd8c1b1593f/json/
never set https://us.sentry.io/api/0/projects/sentry-sdks/sentry-capacitor/events/8f790964c1984a7bab074aa908100bb1/json/

example/ionic-angular-v7/ios/App/App/capacitor.config.json
```diff
@@ -8,9 +8,6 @@
 	},
 	"webDir": "www",
 	"packageClassList": [
-		"SentryRRWebEvent",
-		"kSentryTransactionNameSourceCustom",
-		"kSentryLevelNone",
-		"SentryCapacitor"
+		"SentryCapacitorPlugin"
```
antonis: Q: Is this change related to the fix?

package.json
```diff
@@ -5,7 +5,7 @@
     "type": "git",
     "url": "https://github.com/getsentry/sentry-capacitor"
   },
-  "version": "2.1.0",
+  "version": "2.1.3",
```
antonis: Q: Is the version bump temporary for testing purposes?

Lms24 (approved):
Thanks for fixing this! Looks good to me!

I'm thinking that the SDK might still set `user.ip_address: '{{auto}}'` in case `setDefaultPii: true`. In case it does, we could still still clean this up but it's not the end of the world either way. `sdk.settings.infer_ip` overrules the other field in both cases.

package.json
lucas-zimerman: thanks for catching this, it was a last minute change because ionic wasn't catching the right version, despite being on node modules

example/ionic-angular-v7/ios/App/App/capacitor.config.json
lucas-zimerman: it's just garbage on the sample app that got fixed by the new spm package, I will roll back this and update the samples on another pr

lucas-zimerman: to not pollute the PR
```suggestion
		"SentryRRWebEvent",
		"kSentryTransactionNameSourceCustom",
		"kSentryLevelNone",
		"SentryCapacitor"
```

package.json
lucas-zimerman:
```suggestion
  "version": "2.1.0",
```

PR #961, opened 2025-08-25
Ref: Refer to SPM Sentry Capacitor on Sample apps

lucas-zimerman:
Since Sentry Capacitor supports SPM, we no longer need to rely on auto generated SPM packages.
This PR clears up the previously auto generated files to use the correct SPM package for Sentry Capacitor.

#skip-changelog.


antonis (approved):
LGTM 🚀

Issue #979, opened 2025-09-16
Broken build on iOS with 2.3.0


pfumagalli:
### Environment

All fairly standard...

### Steps to Reproduce

During a normal build I get:

```
[21:33:03]: ▸ ::error file=/Users/runner/work/myapp/myapp/node_modules/@sentry/capacitor/ios/Sources/SentryCapacitorPlugin/SentryCapacitorPlugin.swift,line=73,col=50::argument passed to call that takes no arguments
[21:33:03]: ▸             let options = try Options.init(dict: optionsDict)
[21:33:03]: ▸                                           ~~~~~~~^~~~~~~~~~~~
```

### Expected Result

The new version of Sentry/HybridSDK (8.55.0) seems to be dropping that `init` signature for `Options`. Reverting to `@sentry/capacitor` 2.2.0 (which pulls in Sentry/HybridSDK 8.53.2) fixes the problem

pfumagalli:
Thanks, works!

PR #981, opened 2025-09-17
Test Native Build

lucas-zimerman:
Currently, we don't validate if the SDK will build on the Native SDKs.
This PR ensures the successful build of Sentry Capacitor with our samples on iOS and Android.

The goal is to avoid issues like https://github.com/getsentry/sentry-capacitor/issues/979 in the future
#skip-changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,17 @@
 > [migration guide](https://docs.sentry.io/platforms/javascript/guides/capacitor/migration/) first.
 <!-- prettier-ignore-end -->
 
+## 2.3.1
+
+### Fixes
+
+- iOS build error due to recent changes with Native Sentry Options ([#980](https://github.com/getsentry/sentry-capacitor/pull/980))
+
 ## 2.3.0
+
+> [!WARNING]
+> This release contains an issue where iOS builds will fail, please opt for a lower version or version 2.3.1 or newer.
+> See issue [#979](https://github.com/getsentry/sentry-capacitor/issues/979) for more details.
 
 ### Features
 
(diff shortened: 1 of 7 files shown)


lucas-zimerman:
Same PR pointing to the problematic release could have detected the issue: https://github.com/getsentry/sentry-capacitor/actions/runs/17809889217/job/50630853723

example/ionic-vue3/yarn.lock
```diff
@@ -1588,7 +1598,7 @@
     unplugin "1.0.1"
 
 "@sentry/capacitor@file:.yalc/@sentry/capacitor":
-  version "2.2.0"
+  version "2.3.0"
```
antonis, line 1601: q: Shouldn't this point to the latest `2.3.1`?

.github/workflows/sample-build.yml
```diff
@@ -0,0 +1,130 @@
...
+          GRADLE_OPTS: -Dorg.gradle.daemon=false
+        run: ./gradlew --no-daemon assembleDebug
+
+      - name: Upload Android APK
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ matrix.bump }}-android-apk
+          if-no-files-found: warn
+          path: |
+            ${{ matrix.SAMPLE_DIR }}/android/app/build/outputs/apk/debug/*.apk
```
antonis, line 105: q: Should we upload the logs for Android too instead of the apk?

antonis (approved):
The changes LGTM and the [builds](https://github.com/getsentry/sentry-capacitor/actions/runs/17825199191/job/50676769184?pr=981) are 🟢 🎸

.github/workflows/sample-build.yml
lucas-zimerman: for the time being, we can rely on the `gradlew` log from the action.

example/ionic-vue3/yarn.lock
lucas-zimerman: It shouldn't impact since we don't check if the lock file from samples are dirty, and capacitor version can always change here. I'd say we can ignore it.

PR #995, opened 2025-10-07
Add Security vulnerability warning to sample apps.

lucas-zimerman:
Sample apps uses old dependencies that may contain security vulnerabilities. Those sample apps are only used for demonstration purpose and Integration tests.

This only adds a warning that those sample apps shouldn't be used in production.

If any security vulnerability exists on the sample apps, they will in any way affect the end user SDK of Sentry Capacitor.

Based on https://github.com/getsentry/sentry-cordova/pull/386


antonis (approved):
LGTM 🦸

PR #996, opened 2025-10-07
Fix security/dependabot/589

lucas-zimerman:
Lets use a tmp dependency that uses a version without any security warnings.

#skip-changelog


antonis (approved):
LGTM since the earliest fixed version is 0.2.4 🚀

PR #999, opened 2025-10-13
Update CHANGELOG.md with iOS 26 mention

lucas-zimerman:
By bumping to the latest SDK of Sentry Cocoa we also add support to iOS 26.

This PR only adds a new mention to support to iOS 26 so users know when support was added to Sentry Capacitor.

#skip-changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ## Unreleased
 
-## Fixes
+## Feature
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,10 @@
 
 ## Unreleased
 
+## Fixes
```
lucas-zimerman:
```suggestion
## Feature
```

PR #1014, opened 2025-10-21
fix: do not hardcode the iphone version to be tested.

lucas-zimerman:
By hard-coding the iPhone version, we have the occasional error where the iPhone is not available on future versions.
This fix allows us to always select the latest available iPhone from Github VM.

#skip-changelog.


antonis (approved):
LGTM! That's a nice enhancement 🏅 
I wonder if if it would affect test stability and if it makes sense to try this in RN too.

PR #1086, opened 2026-01-20
Ref(V3): Bump sentry cocoa v9

lucas-zimerman:
This PR Bumps Sentry Cocoa V8 to V9, this brings a break change that the minimum supported ios version is 15.0.

### DIFF
--- a/.github/workflows/danger.yml
+++ b/.github/workflows/danger.yml
@@ -7,6 +7,7 @@ on:
 permissions:
   contents: write
   pull-requests: write
+  statuses: write # Required for the danger action to commit a message on the PR.
 jobs:
   danger:
     runs-on: ubuntu-latest


.github/workflows/danger.yml
```diff
@@ -7,6 +7,7 @@ on:
 permissions:
   contents: write
   pull-requests: write
+  statuses: write # Required for the danger action to commit a message on the PR.
```
lucas-zimerman, line 10: Something broke danger and this is didn't fix it.
```suggestion
```

antonis: Is this comment/suggestion still valid? Danger is currently 🟢

lucas-zimerman: yeah its broken but still giving 🟢 for some reason

lucas-zimerman: i ll check with the danger experts about it

PR #1088, opened 2026-01-21
Fix(V3): Duplicated session / Fix incompatibility with Cocoa V9

lucas-zimerman:
## Dublicated Session

Due to not replacing the `defaultIntegration`, we have a side effect that the sibling SDK will [add](https://github.com/getsentry/sentry-javascript/blob/80b4705eb46cb43212a61d73bf3207e27b750eff/packages/browser/src/sdk.ts#L40) their own integrations, inserting the `BrowserSession` automatically. This is wrong since both native and the JS implementation has auto session implementation.

The correct fix was applied by adding the capacitor integrations into `defaultIntegrations`, overwriting the sibling SDK integrations and thus, only having a single session being created.

A minor refactor was also implemented where I renamed `finalOption` to `sharedOption` since it matches what it is actually being used, as a shared data instead of the final option being used.

Before:
image

After: 
image

image

## Cocoa Fix

It just appeared now that `SentryOptionsInternal` is not available on SPM, so I just opted for just insert the logic locally .

### DIFF
--- a/ios/Sources/SentryCapacitorPlugin/SentryCapacitorPlugin.swift
+++ b/ios/Sources/SentryCapacitorPlugin/SentryCapacitorPlugin.swift
@@ -133,8 +133,8 @@ public class SentryCapacitorPlugin: CAPPlugin, CAPBridgedPlugin {
             options.sessionTrackingIntervalMillis = UInt(sessionTrackingIntervalMillis)
         }
 
-        if let maxBreadcrumbs = dict["maxBreadcrumbs"] as? UInt {
-            options.maxBreadcrumbs = maxBreadcrumbs
+        if let maxBreadcrumbs = dict["maxBreadcrumbs"] as? Int {
+            options.maxBreadcrumbs = UInt(maxBreadcrumbs)
         }
 
         if let enableNativeCrashHandling = dict["enableNativeCrashHandling"] as? Bool {


ios/Sources/SentryCapacitorPlugin/SentryCapacitorPlugin.swift
```diff
@@ -102,8 +97,67 @@ public class SentryCapacitorPlugin: CAPPlugin, CAPBridgedPlugin {
...
+            options.enableAutoSessionTracking = enableAutoSessionTracking
         }
+
+        if let sessionTrackingIntervalMillis = dict["sessionTrackingIntervalMillis"] as? Int {
+            options.sessionTrackingIntervalMillis = UInt(sessionTrackingIntervalMillis)
+        }
+
+        if let maxBreadcrumbs = dict["maxBreadcrumbs"] as? Int {
+            options.maxBreadcrumbs = UInt(maxBreadcrumbs)
+        }
```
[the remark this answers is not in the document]
lucas-zimerman: I'd rather limit this on the JavaScript side.
The PR is not changing the behavior of `sessionTrackingIntervalMillis` and `maxBreadcrumbs` so lets leave it as is and fix it on a separated PR.

CHANGELOG.md
```diff
@@ -6,6 +6,18 @@
...
 
+## Unreleased
+
+### Fixes
+
+- Duplicated session When running Capacitor as an app ([#1088](https://github.com/getsentry/sentry-capacitor/pull/1088))
+
+### Dependencies
+
+- Bump Cocoa SDK from v8.56.2 to v9.2.0 ([#1088](https://github.com/getsentry/sentry-capacitor/pull/1088), [#1086](https://github.com/getsentry/sentry-capacitor/pull/1086))
```
antonis, line 17: We've already bumped to v9 with https://github.com/getsentry/sentry-capacitor/pull/1086, correct?

antonis: Ok. It's just a changelog addition :)

lucas-zimerman: I missed that because danger was broken :( Opened a PR to sort it out https://github.com/getsentry/sentry-capacitor/pull/1094

PR #1072, opened 2026-01-09
ci(release): Switch from action-prepare-release to Craft

BYK:
## Summary

This PR migrates from the deprecated `action-prepare-release` to the new Craft GitHub Actions.

## Changes

- Migrated `.github/workflows/release.yml` to Craft reusable workflow

## Documentation

See https://getsentry.github.io/craft/github-actions/ for more information.

#skip-changelog

PR #1094, opened 2026-01-22
Fix: Danger not working

lucas-zimerman:
Previous runs were throwing :green_circle: on CI, despite not having a changelog. 
The issue was introduced by https://github.com/getsentry/sentry-capacitor/pull/1072/files#diff-9ed6d3c8f52cf7b6862dd9628741eda3dbcd3683b1485977c483b6571e989904
Where it complains about 
```
Could not add a commit status, the GitHub token for Danger does not have access rights.
If the build fails, then danger will use a failing exit code.
```

A 'fix' was introduced by adding `statuses` but that only made the PR to always pass despite the missing changelog.

Without the permissions, the danger action works correctly.

Question: Do we really need the permissions here? If so, do we need to update the token on this repo?

### DIFF
--- a/.github/workflows/danger.yml
+++ b/.github/workflows/danger.yml
@@ -3,9 +3,6 @@ on:
   pull_request:
     types: [opened, synchronize, reopened, edited, ready_for_review, labeled, unlabeled]
 
-permissions:
-  contents: write
-  pull-requests: write
 jobs:
   danger:
     runs-on: ubuntu-latest


.github/workflows/danger.yml
```diff
@@ -3,13 +3,13 @@ on:
   pull_request:
     types: [opened, synchronize, reopened, edited, ready_for_review, labeled, unlabeled]
 
-
 permissions:
   contents: write
   pull-requests: write
```
lucas-zimerman:
```suggestion
  pull-requests: read
```

.github/workflows/danger.yml
```diff
@@ -3,13 +3,13 @@ on:
   pull_request:
     types: [opened, synchronize, reopened, edited, ready_for_review, labeled, unlabeled]
 
-
 permissions:
   contents: write
-  pull-requests: write
-  statuses: write # Required for the danger action to commit a message on the PR.
+  pull-requests: read
```
lucas-zimerman:
```suggestion
  pull-requests: write
```

.github/workflows/danger.yml
```diff
@@ -3,13 +3,13 @@ on:
...
   contents: write
   pull-requests: write
-  statuses: write # Required for the danger action to commit a message on the PR.
 jobs:
   danger:
     runs-on: ubuntu-latest
     steps:
       - uses: getsentry/github-workflows/danger@17cc15eb58ea3687cd8f2714a4192dcee4aa09ef
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
lucas-zimerman:
```suggestion
```

BYK:
If you are using automated changelogs with craft, you should not even need danger at this point? getsentry/sentry-wizard and getsentry/sentry-python switched to that

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@sentry/capacitor",
+  "name": "@sentry/capacitor-blabla",
```
lucas-zimerman:
```suggestion
  "name": "@sentry/capacitor",
```

PR #1123, opened 2026-02-13
Fix Android logcat spam from empty call.resolve()

mattriese:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Wrap all `call.resolve()` invocations with `JSObject()` (e.g. `call.resolve(new JSObject())`) for plugin methods that return no meaningful data, instead of calling `call.resolve()` with no argument.

This prevents Android logcat from being flooded with blank/undefined console log entries. When Capacitor's bridge completes a plugin call, it logs the resolved value to logcat. Calling `call.resolve()` with no argument causes Capacitor to log `"File: - Line 333 - Msg: undefined"` (or similar) for every completed call. By passing an empty `JSObject` instead, the bridge receives a valid serializable value and does not emit these noisy undefined logs.

This addresses a known Capacitor Android bug: [ionic-team/capacitor#8063](https://github.com/ionic-team/capacitor/issues/8063) — "PluginCall.resolve logs in Logcat inconsistently". When `call.resolve()` is invoked with no argument, the Capacitor Console API logs `undefined` to logcat for every completed plugin call.

The Sentry Capacitor plugin makes many bridge calls, and several methods (especially `addBreadcrumb`, which runs for every console log captured as a breadcrumb) resolve with no return value. In apps with frequent logging or breadcrumb activity, this produces significant logcat spam—hundreds of lines like:

```
Capacitor/Console I  File: -  Line: 333  Msg: undefined
```

This obscures real logs and makes Android debugging difficult. The fix is a simple workaround: pass `new JSObject()` to `call.resolve()` instead of nothing. Capacitor does not log when given a proper object, so the spam is eliminated.

Related: [ionic-team/capacitor#8063](https://github.com/ionic-team/capacitor/issues/8063) (closed, fixed in Capacitor core via [#8225](https://github.com/ionic-team/capacitor/pull/8225)). This change provides the same workaround for users on older Capacitor versions and ensures clean logcat output regardless.

I have been using this fork in my app for 2+ months now and it has been working great. I have not noticed any regressions. I forked off of 2.4.1 and just merged in 3.0.0. I will mark this as ready after using the new merged version for a while. 

1. Before the fix: Run the app, trigger normal Sentry activity (console breadcrumbs, init, etc.). Observe logcat: many lines of `Capacitor/Console I ... Msg: undefined`.
2. After the fix: Same scenario. Logcat no longer shows undefined entries for Sentry plugin completions; only meaningful logs appear.

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

None. This is a self-contained bugfix.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "@sentry/sentry-capacitor",
+  "name": "@sentry/capacitor",
   "homepage": "https://github.com/mattriese/sentry-capacitor",
   "repository": {
     "type": "git",


.gitignore
```diff
@@ -3,7 +3,6 @@
 # And the CLI then renames it
 
 # node files
-dist/
```
lucas-zimerman, line 6: Lets keep dist on this list

package.json
```diff
@@ -1,11 +1,11 @@
 {
   "name": "@sentry/capacitor",
-  "homepage": "https://github.com/getsentry/sentry-capacitor",
+  "homepage": "https://github.com/mattriese/sentry-capacitor",
```
lucas-zimerman, line 3: Would you be able to undo the changes here?

android/src/main/java/io/sentry/capacitor/SentryCapacitor.java
```diff
@@ -266,7 +239,9 @@ public void captureEnvelope(PluginCall call) {
             call.reject(errorMsg);
             return;
         }
-        call.resolve();
+        // Resolve with empty object to avoid Capacitor logging "undefined" to Logcat
+        // See: https://github.com/ionic-team/capacitor/issues/4986
```
lucas-zimerman, line 243: Is this link correct? it redirects to a PR https://github.com/ionic-team/capacitor/pull/4986

PR #1106, opened 2026-02-05
Bump: Sentry cocoa to 9.3.0

lucas-zimerman:
* Bump Sentry cocoa to 9.3.0 https://github.com/getsentry/sentry-cocoa/releases/tag/9.3.0

Tested with CI / Manual

PR #1277, opened 2026-06-10
fix(scripts): match podspec dependency name in update-cocoa.sh

lucas-zimerman:
## Summary
- The Cocoa SDK auto-bump workflow was failing with "Failed to find sentry-cocoa version in either ... SentryCapacitor.podspec".
- PR #1106 renamed the podspec dependency from `'Sentry/HybridSDK'` to `'Sentry'`, but `scripts/update-cocoa.sh`'s regex was never updated to match, silently breaking `get-version`/`set-version` for the auto-updater.
- Updated the regex in both `get-version` and `set-version` to match `'Sentry'`.

## Test plan
- [x] `./scripts/update-cocoa.sh get-version` now returns `9.16.1` (matches `SentryCapacitor.podspec`)

### DIFF
--- a/scripts/update-cocoa.sh
+++ b/scripts/update-cocoa.sh
@@ -8,7 +8,7 @@ case $1 in
 get-version)
     if [[ -f "$podFile" ]]; then
         podContent=$(cat $podFile)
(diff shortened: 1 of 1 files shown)


antonis (approved):
LGTM

PR #1320, opened 2026-07-06
docs: fix typos in code comments

Avicennasis:
## Summary

Fix 4 issues in source code comments and error messages:

- `src/options.ts`: "at the cost of lost on" → "at the cost of loss of"
- `src/options.ts`: "also enabled HTTP Errors" → "also enables HTTP Errors"
- `src/options.ts`: "On the future" → "In the future" (2 occurrences)
- `src/client.ts`: "integraiton" → "integration"

No functional changes — comments and error messages only.


lucas-zimerman (approved):

PR #1350, opened 2026-07-29
fix(ios): honor enableAppHangTracking, appHangTimeoutInterval and enableWatchdogTerminationTracking

### LINKED ISSUE #1326: iOS bridge silently drops enableAppHangTracking / appHangTimeoutInterval (and enableWatchdogTerminationTracking) passed to Sentry.init
### Environment

* `@sentry/capacitor` 4.2.0 (Capacitor 8, iOS)

### Description

The JS layer deliberately forwards the App Hang options to the native layer on iOS — `FilterNativeOptions()` in `dist/esm/nativeOptions.js`:

```js
function iOSParameters(options) {
    return Capacitor.getPlatform() === 'ios'
        ? {
            enableAppHangTracking: options.enableAppHangTracking,
            appHangTimeoutInterval: options.appHangTimeoutInterval,
        }
        : {};
}
```

…but the Swift side never reads them. `createOptions(from:)` in `ios/Sources/SentryCapacitorPlugin/SentryCapacitorPlugin.swift` maps an explicit allowlist of keys (`dsn`, `debug`, `environment`, `release`, `dist`, `enableAutoSessionTracking`, `sessionTrackingIntervalMillis`, `maxBreadcrumbs`, `enableNativeCrashHandling`, `attachStacktrace`, `sampleRate`, `tracesSampleRate`, `enableAutoPerformanceTracing`, `strictTraceContinuation`, `orgId`) and silently discards everything else.

As a result, passing `enableAppHangTracking: false` or a custom `appHangTimeoutInterval` to `Sentry.init()` has no effect: sentry-cocoa starts with its defaults (App Hang tracking enabled, 2 s timeout), with no warning anywhere.

`enableWatchdogTerminationTracking` appears to be affected the same way: forwarded by `FilterNativeOptions()`, never parsed by `createOptions(from:)`.

### Steps to reproduce

1. `Sentry.init({ dsn, enableAppHangTracking: false })` (or `appHangTimeoutInterval: 10`) in a Capacitor iOS app.
2. Observe that App Hang events are still produced with the default 2 s threshold.

### Expected

Either the options are honored by the native bridge, or they are rejected/documented as unsupported instead of being silently dropped.

### Context

We hit this while mitigating false-positive `App Hang Fully Blocked` events from a suspended app (same family as getsentry/sentry-cocoa#8317). The only working lever from the JS layer today is calling `pauseAppHangTracking()` after init, since the `initNativeSdk` path cannot disable the integration.

lucas-zimerman:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

`createOptions(from:)` in `ios/Sources/SentryCapacitorPlugin/SentryCapacitorPlugin.swift` only mapped an explicit allowlist of keys onto the native `Options` object. `enableAppHangTracking`, `appHangTimeoutInterval`, and `enableWatchdogTerminationTracking` were not in that allowlist, so they were silently dropped even though the JS layer's `FilterNativeOptions()` already forwarded them on iOS. This PR adds the three missing keys to `createOptions(from:)` so they are applied to the native `Options` object before `SentrySDK.start(options:)` is called.

Passing `enableAppHangTracking: false` or a custom `appHangTimeoutInterval` to `Sentry.init()` had no effect on iOS: sentry-cocoa always started with its defaults (App Hang tracking enabled, 2s timeout), with no warning. `enableWatchdogTerminationTracking` was affected the same way.

Fixes #1326

Manual code review against sentry-cocoa 9.16.1's `Options.swift` (the version pinned in `SentryCapacitor.podspec`) to confirm the property names/types (`enableAppHangTracking: Bool`, `appHangTimeoutInterval: TimeInterval`, `enableWatchdogTerminationTracking: Bool`) match, and by mirroring the existing pattern used for every other option in `createOptions(from:)`. I did not have Xcode available to run a full native build/test on this machine.

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
Would be good to add a unit test around `SentryCapacitorPlugin.createOptions(from:)` (currently untested) to lock this in and cover future option additions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@
 > [migration guide](https://docs.sentry.io/platforms/javascript/guides/capacitor/migration/) first.
 <!-- prettier-ignore-end -->
 
+## Unreleased
+
+### Fixes
+
+- iOS: `enableAppHangTracking`, `appHangTimeoutInterval` and `enableWatchdogTerminationTracking` are now correctly passed into Sentry Cocoa SDK ([#1326](https://github.com/getsentry/sentry-capacitor/issues/1326))
+
 ## 4.2.0
 
 ### Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 ### Fixes
 
-- iOS: `enableAppHangTracking`, `appHangTimeoutInterval` and `enableWatchdogTerminationTracking` are now correctly passed into Sentry Cocoa SDK ([#1326](https://github.com/getsentry/sentry-capacitor/issues/1326))
+- iOS: `enableAppHangTracking`, `appHangTimeoutInterval` and `enableWatchdogTerminationTracking` are now correctly passed into Sentry Cocoa SDK ([#1350](https://github.com/getsentry/sentry-capacitor/pull/1350))
 
 ## 4.2.0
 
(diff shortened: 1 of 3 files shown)


antonis (approved):
LGTM

PR #1384, opened 2026-09-02
fix(ios): forward enableCaptureFailedRequests and sendDefaultPii to Sentry Cocoa

### LINKED ISSUE #1377: iOS: enableCaptureFailedRequests is silently ignored — createOptions() allow-list does not read it (follow-up to #1350)
### Summary

On iOS, `enableCaptureFailedRequests` is documented as a supported option and is forwarded across the bridge by `FilterNativeOptions`, but the native plugin never reads it — so the value is silently discarded and the Cocoa default (`true`) is used regardless.

This is the same class of bug as #1350, which fixed `enableAppHangTracking`, `appHangTimeoutInterval` and `enableWatchdogTerminationTracking` in 4.3.0. `enableCaptureFailedRequests` was not included in that fix and is still dropped on 4.3.0.

### Cause

`ios/Sources/SentryCapacitorPlugin/SentryCapacitorPlugin.swift` → `createOptions(from:)` builds `Options` from a hand-maintained allow-list rather than delegating to `SentryOptions(dict:)`. On 4.3.0 it reads:

```
dsn, debug, environment, release, dist, enableAutoSessionTracking,
sessionTrackingIntervalMillis, maxBreadcrumbs, enableNativeCrashHandling,
attachStacktrace, sampleRate, tracesSampleRate, enableAutoPerformanceTracing,
enableAppHangTracking, appHangTimeoutInterval,
enableWatchdogTerminationTracking, strictTraceContinuation, orgId
```

Meanwhile `src/nativeOptions.ts` `FilterNativeOptions` explicitly forwards `enableCaptureFailedRequests`, and `src/options.ts` documents it. The key crosses the bridge and is then ignored.

`sendDefaultPii` is forwarded by `FilterNativeOptions` and likewise not read by `createOptions`.

### Impact

With `CapacitorHttp: { enabled: true }`, all web-layer HTTP traffic is routed through native `URLSession`, so `SentryNetworkTracker` reports every failed request as an `HTTPClientError` with a pure-native stack. Those events are produced by the native layer and **bypass the JS `beforeSend`**, so no JS-side `beforeSend` / `ignoreErrors` filter can drop them, and `enableCaptureFailedRequests: false` — the documented lever — has no effect.

In our app this produced ~30k events / ~11.8k users on a single issue before we traced it. Because the gap is in compiled Swift, it is not fixable by an over-the-air JS update; it required a new native App Store build.

### Reproduction

1. `Sentry.init({ dsn, enableCaptureFailedRequests: false })` in a Capacitor iOS app on 4.3.0.
2. Trigger any request that fails with a 5xx.
3. An `HTTPClientError` event is still reported, with `culprit` `-[__NSCFLocalSessionTask _task_onqueue_didFinish]`.

Confirmable without a debug build — the resolved options are visible in any event's `sdk` block:

```json
"sdk": {
  "name": "sentry.cocoa.capacitor",
  "version": "9.16.1",
  "integrations": ["Crash","PerformanceTracking","WatchdogTerminationTracking",
                   "NetworkTracking","ANRTracking", "..."],
  "features": ["captureFailedRequests","experimentalViewRenderer","dataSwizzling","metrics"]
}
```

`captureFailedRequests` in `features` means the native SDK resolved `enableCaptureFailedRequests` to `true` despite `false` being passed.

### Suggested fix

Add the key alongside the ones #1350 introduced:

```swift
if let enableCaptureFailedRequests = dict["enableCaptureFailedRequests"] as? Bool {
    options.enableCaptureFailedRequests = enableCaptureFailedRequests
}
```

More durably: have `createOptions(from:)` delegate to `SentryOptions(dict:)` so the plugin cannot drift from the Cocoa option set again. The hand-maintained allow-list has now silently dropped four documented options; a key added to `FilterNativeOptions` without a matching Swift branch fails silently, with no warning at init and no way for an app developer to tell the option was ignored.

At minimum, logging unrecognised keys when `debug: true` would make this discoverable.

### Version

- `@sentry/capacitor` 4.3.0 (also reproduces on 4.1.0)
- `sentry-cocoa` 9.16.1
- Capacitor 8

lucas-zimerman:
- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
`createOptions(from:)` in `ios/Sources/SentryCapacitorPlugin/SentryCapacitorPlugin.swift` builds `Options` from a hand-maintained allow-list of dictionary keys rather than delegating to `SentryOptions(dict:)`. Two documented, JS-bridged options were missing from that list and were silently dropped:

- `enableCaptureFailedRequests`
- `sendDefaultPii`

Both are added to the allow-list, following the same pattern used for the other bridged boolean options.

On iOS, `enableCaptureFailedRequests` is documented and forwarded across the bridge by `FilterNativeOptions`, but the native plugin never read it, so the Cocoa default (`true`) was always used regardless of what the app configured. `sendDefaultPii` had the same gap.

This is the same class of bug as #1350, which fixed `enableAppHangTracking`, `appHangTimeoutInterval` and `enableWatchdogTerminationTracking` in 4.3.0 — `enableCaptureFailedRequests` and `sendDefaultPii` were not included in that fix.

With `CapacitorHttp: { enabled: true }`, HTTP traffic is routed through native `URLSession`, so `SentryNetworkTracker` reports failed requests as `HTTPClientError` events with a pure-native stack that bypass the JS `beforeSend`. Since `enableCaptureFailedRequests: false` had no effect, those events could not be filtered client-side.

Fixes #1377

Verified by reading the Sentry Cocoa `SentryOptions.h` header to confirm both `enableCaptureFailedRequests` and `sendDefaultPii` are `BOOL` properties on `SentryOptions`, and that the added code matches the existing bridging pattern in `createOptions(from:)`. No Swift toolchain available in this environment to build/run the iOS test suite locally — deferring to CI.

## :pencil: Checklist
- [x] I updated the docs if needed.
- [x] I updated the wizard if needed.
- [ ] All tests passing
- [x] Public API changes reviewed by another Mobile SDK team member or implemented according to the [develop docs](https://develop.sentry.dev/) spec
- [x] No breaking changes

## :crystal_ball: Next steps
Longer term, `createOptions(from:)` could delegate to `SentryOptions(dict:)` so the plugin can't silently drift from the Cocoa option set again, as suggested in #1377.


antonis (approved):
LGTM

